Issue #112, opened 2021-03-12
Comments for Image Api Issue


sumitra19jha:
This rest end point needs to be provided with a proper documentation. It leads to the following error
![Screenshot 2021-03-12 at 6 00 41 AM](https://user-images.githubusercontent.com/35868598/110873900-b4b20d00-82f8-11eb-8f66-70915168a37e.png)

The issue has been provided with a solution [here](https://stackoverflow.com/questions/43260643/how-to-resolve-nodejs-error-enoent-no-such-file-or-directory).
A proper documentation needs to be provided to provide the user information with adding the `images` folder to prevent the issue.

sumitra19jha:
Should I make changes for the documentation of the api?

PR #113, opened 2021-03-12
User Pagination Query and Comment for image api

sumitra19jha:
comments added #112


sumitra19jha:
Pagination for the users.
The query is being made flexible to handle different conditions.

PR #124, opened 2021-03-12
added comments and put try catch blocks

yasharth291:


yasharth291:
@heyjordn  please look into this as well

Issue #192, opened 2021-03-30
Upgrade node-notifier package to > version 8.0.1 in package-lock.json


palisadoes:
We need to remedy the following known vulnerability: https://github.com/advisories/GHSA-5fw9-fq32-wv5p

GitHub cannot automatically do the required update to > 8.0.1 because:

```
The latest possible version that can be installed is 7.0.2 because of the following conflicting dependency:

jest@26.0.1 requires node-notifier@^7.0.0 via a transitive dependency on @jest/reporters@26.0.1
The earliest fixed version is 8.0.1.
```

This update needs to be implemented and tested.

palisadoes:
Assigned to @Hrishikesh-Bhagwat

somya339:
I want to work on this issue.

PR #196, opened 2021-04-01
Create package-lock.json

somya339:
node-notifier and netmask versions are updated from 7.0.2 to 8.0.1 and 1.0.2 to 2.0.1 respectively.
Referencing issue #192 and #191 .


somya339:
The fork was not updated leading to conflicting commits.

PR #199, opened 2021-04-01
Fixed the problem of Netsapp ^1.0.6 Dependency

### LINKED ISSUE #191: Update package-lock.json to upgrade to netmask version 2.0.1
There is a high severity vulnerability in the netmask package. Details can be found here https://github.com/advisories/GHSA-pch5-whg9-qr2r

The solution is to upgrade netmask to version 2.0.1 or later. There are dependencies that will need to be remedied.

```
The latest possible version that can be installed is 1.0.6 because of the following conflicting dependency:

pm2@4.5.5 requires netmask@^1.0.6 via a transitive dependency on pac-resolver@3.0.0
```

This dependency needs to be fixed so that we can upgrade to netmask version 2.0.1

yasharth291:

Dependencies

**Summary**

Fixes #191 

Fixed the high risked problem of the netsapp


yasharth291:
thanks sir

PR #204, opened 2021-04-02
1. Search users By their first name or last name, 2. Now Any user can save a post of any organization 

XshubhamX:
I have build a Query Which will help in Searching User By their first names or last names

This query will help when users want to search other users ,so if they just know their names they can query data by passing name in graphql query. Now in frontend if there is a search option then at every stroke a query should me made to the api sending the query string and user will get the matching names of users.

I have also edited the user model and added a new field in it which is an array that will contain ids of posts which a user want to save. 

I hope this is useful 👍🏻

### DIFF
--- a/db.js
+++ b/db.js
@@ -1,4 +1,5 @@
 const mongoose = require('mongoose');
+const chalk = require('chalk');
 
 const connect = async () => {
   try {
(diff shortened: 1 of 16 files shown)


XshubhamX:
@Laurell876 @Sagar2366 @palisadoes Sir please have a look at my pull requests, are they reliable.

XshubhamX:
@Sagar2366 Sir can you please have a look at this pull request ,if its ok then I can start further process

XshubhamX:
@Sagar2366  Sir should I do more work on this like I will create a query which will fetch all the posts saved by the user , or should I stop working on this .Please tell sir.

PR #209, opened 2021-04-04
removing redundant file operations in case of invalid extension

anonymous-ME:

Refactoring image upload and extension check util.

No.

Not required.

**Summary**

The image upload and image extension check function has been refactored to avoid redundant file operations in case of invalid extension, Its just that now we are avoiding writing to the disk if we have invalid file extension instead of first writing and then deleting the file.

Also, I've removed delete function call from extension validator as it should not be responsible for deleting the files.

No. Essentially the end result will be the same.


palisadoes:
@anonymous-ME Please resolve the merge conflicts

anonymous-ME:
@palisadoes conflicts are now resolved.

PR #223, opened 2021-04-08
Resolves #210 Image upload mutations

XshubhamX:
I have created Image upload test for signup mutation by sending request in multipart form data. @Sagar2366  Sir.

PR #240, opened 2021-05-17
bug issue#223 resolved

ssaumyaa7:
-bugfix 

**Summary**
-There was bug while comparing, string and int cannot be === as there needs to be the same datatype but when we use == the datatypes are ignored rather the content is compared.

**Other information**
issue no. #223

### DIFF
--- a/resolvers/functions/creatorCheck.js
+++ b/resolvers/functions/creatorCheck.js
@@ -1,5 +1,5 @@
 const creatorCheck = (context, org) => {
-  const isCreator = org.creator === context.userId;
+  const isCreator = org.creator == context.userId;
   if (!isCreator) {
     throw new Error("Users cannot delete organizations they didn't create");
   }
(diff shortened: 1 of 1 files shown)


resolvers/functions/creatorCheck.js
```diff
@@ -1,5 +1,5 @@
 const creatorCheck = (context, org) => {
-  const isCreator = org.creator === context.userId;
+  const isCreator = org.creator == context.userId;
   if (!isCreator) {
```
impiyush83, line 3: == is loosely check. which is a bad practice
use === ,  match types of the variables

impiyush83 (changes requested):
Use === instead of ==, by checking datatypes

ssaumyaa7:
@impiyush83 taking your point in consideration I'll make the necessary changes again.... Thankyou

ssaumyaa7:
@impiyush83 @palisadoes Rather than using == while comparing, converted type to string

impiyush83 (approved):

palisadoes:
@ssaumyaa7  @impiyush83 Let's have more meaningful pull requests. For example using `===` consistently in all dart files would be a more useful, and require less interruptions for approvals. Please consider the time of the reviewers.

ssaumyaa7:
@ranilwallace sir, @DMills27 sir, 
Please review, even after getting this merged. This is not reviewed hence not reflects my name in the contribution

PR #318, opened 2021-06-24
Added yarn package since it is faster than npm

saymedgm:
I have added yarn package since it is faster than npm which will help project to run faster

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,8 @@
         "validator": "^13.0.0",
         "winston": "^3.3.3",
         "ws": ">=7.4.6",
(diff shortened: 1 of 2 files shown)


Sagar2366:
@saymedgm thanks for raising the PR.
Please use issue and PR templates to create new issues or PR and provide as much as possible.

Do we need to push package-lock.json ? Will it auto-generate on compiling project ?

saymedgm:
Yes @Sagar2366 I used npm install yarn so we need to push package-lock.json

Sagar2366 (approved):

PR #337, opened 2021-08-07
Fixed Create Event Mutation: Fixes #276

### LINKED ISSUE #276: Only members of organisations should be allowed to create events for them
**Describe the bug**
Any user can create events for any organisation

**To Reproduce**
1. create a new organisation
2. login with a diff account
3. use create-event mutation without joining the org 

**Expected behavior**
### User should only be able to create events in the organisation if they are a member of it.
**Actual behavior**
Anyone can create events.

**Additional details**
The logic in the [/resolvers/event_mutations/createEvent.js](https://github.com/PalisadoesFoundation/talawa-api/blob/master/resolvers/event_mutations/createEvent.js) is to be changed to deal with this.

uzair-ali10:

BugFix

**Issue Number:**

Fixes #276 

No

**Snapshots/Videos:**

Screenshot from 2021-08-08 00-56-48

The above image shows a user who isn't a part of org trying to create an event in org and received an error 

No

**Summary**

Previously a user was able to create an event in an org even if they aren't a member of the org, This PR tends to resolve this issue.

Here `flag` is a variable that is used to handle condition. if `flag = 0` that means the user isn't a part of org and the API returns an error. If `flag = 1` this means that the user is a part of the org and we proceed with further steps.

```  
JoinedOrg.forEach((orgID) => {
    if (orgID.equals(args.data.organizationId)) {
      flag = 1;
    }
  });
```
The above function tries to check if the org in which the user is trying to create an event exists in the `JoinedOrg` array which is fetched from DB. If the user is a part of that org then the value of `flag` is changed to `1`

```
    CreatedOrg.forEach((orgID) => {
      if (orgID.equals(args.data.organizationId)) {
        flag = 1;
      }
    });
 ```
The above function tries to check if the org in which the user is trying to create an event exists in the `CreatedOrg` array which is fetched from DB. If the user is a part of that org then the value of `flag` is changed to `1`

No

**Other information**

Na

Yes

### DIFF
--- a/resolvers/event_mutations/createEvent.js
+++ b/resolvers/event_mutations/createEvent.js
@@ -14,7 +14,6 @@ const createEvent = async (parent, args, context) => {
       'user'
     );
   }
-
   // ensure organization exists
   const org = await Organization.findOne({ _id: args.data.organizationId });
   if (!org) {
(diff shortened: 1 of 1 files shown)


ranilwallace (approved):

uzair-ali10:
@DMills27  @Laurell876 

Please Review !!

PR #624, opened 2022-03-18
Fixing issue #588

### LINKED ISSUE #588: Resolvers: Create tests for postsByOrganization.js

Tests need to be written for file `lib/resolvers/post_query/postsByOrganization.js`

- We will need the API to be refactored for all methods, classes and/or functions found in this file for testing to be correctly executed.
- When complete, all all methods, classes and/or functions in the refactored file will need to be tested. These tests must be placed in a
single file with the name `tests/resolvers/post_query/postsByOrganization.spec.js`. You may need to create the appropriate directory structure to do this.

### IMPORTANT:

- https://github.com/PalisadoesFoundation/talawa-api/issues/490

### PR Acceptance Criteria

- When complete this file must show **100%** coverage when merged into the code base. This will be clearly visible when you submit your PR.
- [The current code coverage for the file can be found here](https://codecov.io/gh/PalisadoesFoundation/talawa-api/tree/develop/lib/resolvers/post_query/). If the file isn't found in this directory, or there is a 404 error, then tests have not been created.
- The PR will show a report for the code coverage for the file you have added. You can use that as a guide.

ebrahim354:
### bugfix 

Fixes #588

yes

**Summary**
Fixed a bug found on the postsByOrganization query resolver found after writing tests for the issue #588-

### DIFF
--- a/tests/resolvers/post_query/postsByOrganization.spec.js
+++ b/tests/resolvers/post_query/postsByOrganization.spec.js
@@ -1,0 +1,590 @@
+const axios = require('axios');
+const { URL } = require('../../../constants');
+const getToken = require('../../functions/getToken');
+const shortid = require('shortid');
+
+const database = require('../../../db');
+const getUserId = require('../../functions/getUserId');
+const Organization = require('../../../lib/models/Organization');
+const Post = require('../../../lib/models/Post');
+const Comment = require('../../../lib/models/Comment');
+const User = require('../../../lib/models/User');
+
+// let mainToken;
+// let secondToken;
+// let thirdToken;
+
+let mainUserId;
+let secondUserId;
+let thirdUserId;
+
+let mainOrganization;
+
+let firstPost;
+let secondPost;
+let thirdPost;
+
+beforeAll(async () => {
+  // Initializing a seed data to test on by
+  // creating 3 emails one that holds the main organization
+  // the 2 emails just to like and comment on the testing posts
+  // then creating 3 posts that can be sorted
+  // according to the available criterias.
+
+  try {
+    require('dotenv').config();
+    await database.connect();
+
+    const mainEmail = `${shortid.generate().toLowerCase()}@test.com`;
+    await getToken(mainEmail);
+    mainUserId = await getUserId(mainEmail);
+
+    const secondEmail = `${shortid.generate().toLowerCase()}@test.com`;
+    await getToken(secondEmail);
+    secondUserId = await getUserId(secondEmail);
+
+    const thirdEmail = `${shortid.generate().toLowerCase()}@test.com`;
+    await getToken(thirdEmail);
+    thirdUserId = await getUserId(thirdEmail);
+
+    console.log('created users successfully!');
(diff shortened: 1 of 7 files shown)


tests/resolvers/post_query/postsByOrganization.spec.js
```diff
@@ -0,0 +1,590 @@
...
+const database = require('../../../db');
+const getUserId = require('../../functions/getUserId');
+const Organization = require('../../../lib/models/Organization');
+const Post = require('../../../lib/models/Post');
+const Comment = require('../../../lib/models/Comment');
+const User = require('../../../lib/models/User');
+
+// let mainToken;
+// let secondToken;
+// let thirdToken;
```
JamaicanFriedChicken, line 15: if these variables aren't being used, it would be best to remove them.

ebrahim354: I used them to initialize the seed data using the api then I faced a problem on adding imageUrl so used the database models directly and I'm not sure whether to leave it this way or change it so I just commented them if the database approach is ok with you i will remove them.

ebrahim354: also the sorting problem I just solved it in place but I think a better way is to add a likeCount and commentCount property inside the user document to keep track of them directly from the database but  again I'm just waiting for feed back from you guys.

palisadoes: Thanks, use the count properties.

palisadoes: > I used them to initialize the seed data using the api then I faced a problem on adding imageUrl so instead I used the database models directly and I'm not sure whether to leave it this way or change it so I just commented them if the database approach is ok with you i will remove them.

What do you suggest as the best long term approach?

ebrahim354: > > I used them to initialize the seed data using the api then I faced a problem on adding imageUrl so instead I used the database models directly and I'm not sure whether to leave it this way or change it so I just commented them if the database approach is ok with you i will remove them.
> 
> What do you suggest as the best long term approach?

I think using the database directly is better for performance ( better than using the whole api ) of tests which is beneficial with the increasing of tests added to the system, also the database approach solves the the problem I faced of adding imageUrl field without needing to pass unnecessary input validation etc..

palisadoes: OK, please proceed. We need to fix this as many of our pushes to the repo are intermittently failing.

palisadoes:
OK, please proceed with the DB,  `likeCount` and `commentCount` properties approaches. We need to fix this as many of our pushes to the repo are intermittently failing.

ebrahim354:
> OK, please proceed with the DB, `likeCount` and `commentCount` properties approaches. We need to fix this as many of our pushes to the repo are intermittently failing.

done

palisadoes (approved):

PR #645, opened 2022-03-26
addTest/updateProject.spec.js/issue#570

### LINKED ISSUE #570: Resolvers: Create tests for updateProject.js

Tests need to be written for file `lib/resolvers/event_project_mutations/updateProject.js`

- We will need the API to be refactored for all methods, classes and/or functions found in this file for testing to be correctly executed.
- When complete, all all methods, classes and/or functions in the refactored file will need to be tested. These tests must be placed in a
single file with the name `tests/resolvers/event_project_mutations/updateProject.spec.js`. You may need to create the appropriate directory structure to do this.

### IMPORTANT:

- https://github.com/PalisadoesFoundation/talawa-api/issues/490

### PR Acceptance Criteria

- When complete this file must show **100%** coverage when merged into the code base. This will be clearly visible when you submit your PR.
- [The current code coverage for the file can be found here](https://codecov.io/gh/PalisadoesFoundation/talawa-api/tree/develop/lib/resolvers/event_project_mutations/). If the file isn't found in this directory, or there is a 404 error, then tests have not been created.
- The PR will show a report for the code coverage for the file you have added. You can use that as a guide.

xoldyckk:

Refactors updateProject.js.
Adds test file for updateProject.js

**Issue Number:**

Fixes #570

Yes

**Snapshots/Videos:**

Screenshot from 2022-03-26 11-02-42

No

**Summary**

Refactored some files to throw easily mockable errors when not in production environment. Fixed a bug where comparing mongodb's objectId was implemented incorrectly. Added test file for updateProject.js increasing code coverage across talawa-api project.

No

**Other information**

No

Yes

### DIFF
--- a/tests/resolvers/event_project_mutations/updateProject.spec.js
+++ b/tests/resolvers/event_project_mutations/updateProject.spec.js
@@ -8,6 +8,16 @@ const Organization = require('../../../lib/models/Organization');
 const Event = require('../../../lib/models/Event');
 const EventProject = require('../../../lib/models/EventProject');
 
+let testUser1;
+
+let testUser2;
+
+let testOrganization;
+
+let testEvent;
+
+let testEventProject;
+
 const createUser = async () => {
   const email = `${shortid.generate().toLowerCase()}@test.com`;
   const hashedPassword = await bcrypt.hash('password', 12);
@@ -111,16 +121,6 @@ const createEventProject = async (user, event) => {
 
   return createdEventProject;
 };
-
-let testUser1;
-
-let testUser2;
-
-let testOrganization;
-
-let testEvent;
-
-let testEventProject;
 
 // Read this :- https://jestjs.io/docs/api#beforeallfn-timeout
 beforeAll(async () => {
--- a/tests/resolvers/event_project_mutations/updateProject.spec.js
+++ b/tests/resolvers/event_project_mutations/updateProject.spec.js
@@ -1,0 +1,200 @@
+const mongoose = require('mongoose');
+const bcrypt = require('bcryptjs');
+const shortid = require('shortid');
+const database = require('../../../db');
+const updateProject = require('../../../lib/resolvers/event_project_mutations/updateProject');
+const User = require('../../../lib/models/User');
+const Organization = require('../../../lib/models/Organization');
+const Event = require('../../../lib/models/Event');
+const EventProject = require('../../../lib/models/EventProject');
+
+const createUser = async () => {
+  const email = `${shortid.generate().toLowerCase()}@test.com`;
+  const hashedPassword = await bcrypt.hash('password', 12);
+
+  const newUser = new User({
+    email,
+    password: hashedPassword,
+    firstName: 'firstName',
+    lastName: 'lastName',
+    appLanguageCode: 'en',
+    image: null,
+  });
+
+  const createdUser = await newUser.save();
+
+  return createdUser;
+};
+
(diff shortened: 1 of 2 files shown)


palisadoes:
Can you get the code coverage up to 100%?

xoldyckk:
> Can you get the code coverage up to 100%?

The errors thrown by default in resolvers are not mockable. When mocked they throw `i18n is not initialized, try app.use(i18n.init)` as the error message. This occurs because of the function `requestContext.translate`. This function only runs inside the context of the actual talawa-api server. That means we have to simulate an actual talawa-api server inside our tests to test it. That's why there's this refactor :- 

      process.env.NODE_ENV !== 'production'
        ? 'User not authorized'
        : requestContext.translate('user.notAuthorized')

So, that in testing environment the errors don't call requestContext.translate function. Instead they use a plain string.

And I believe that's what's causing the test coverage to not be 100%.

palisadoes:
Can we mock `requestContext.translate` in some way?

Should we consider running an API server as part of the testing regimen and mock where necessary for the rest? I've written code in the past where I had to spin up an instance of the API to properly test end to end. I realize that this may not be practical in future where running the API may consume all test resources.

xoldyckk:
> Can we mock `requestContext.translate` in some way?
> 
> Should we consider running an API server as part of the testing regimen and mock where necessary for the rest? I've written code in the past where I had to spin up an instance of the API to properly test end to end. I realize that this may not be practical in future where running the API may consume all test resources.


I personally don't have much experience with testing. I only got recently got started with it because of this project. So, my suggestions could be wrong or unreliable. But here's my take:-

In all the integration/e2e test implementations I've seen on testing backends/server, instead of just testing the route/resolver in isolation, they all run tests on the actual server. To make testing in CI runs faster, we can pass the option `--runInBand` to the test script to make it faster. Here's the source :- [https://dev.to/vantanev/make-your-jest-tests-up-to-20-faster-by-changing-a-single-setting-i36](url).

ShwetKhatri2001:
I'm also getting the same problem of not covering the second branch ( involving requestContext.translate ) in my open PRs.
I tried to mock it by initiating i18n with the same settings we have in server file ( `index.js` ) , but didn't get expected behaviour.

But according to me here the scope of our API test case is to check if the error is delivered, not in which language it is delivered. That's why we separated

palisadoes:
@JamaicanFriedChicken @DMills27 This was the case I spoke about today. Please please give your inputs.

palisadoes:
@JamaicanFriedChicken @DMills27 Please respond here so we have a record of your recommended approach.

JamaicanFriedChicken:
@xoldyckk @ShwetKhatri2001 Thanks guys. We will be making this into a separate issue referencing this PR to investigate the cause of this. For now, we will accept the PRs as is, as it relates to i18n even if the test coverage isn't 100%.

tests/resolvers/event_project_mutations/updateProject.spec.js
```diff
@@ -0,0 +1,200 @@
...
+
+let testUser1;
+
+let testUser2;
+
+let testOrganization;
+
+let testEvent;
+
+let testEventProject;
```
JamaicanFriedChicken, line 123: it's best to put these variables at the top of the file.

JamaicanFriedChicken (changes requested):
Still need to make some slight adjustments, almost there.

JamaicanFriedChicken (approved):

PR #648, opened 2022-03-29
resolving issue: #554

ebrahim354:
**What kind of change does this PR introduce?**

bugfix, testing, resolving issue

**Issue Number:**

Fixes #554

**Did you add tests for your changes?**
yes

**Summary**
the previous approach for removing an organization was incomplete and would leave some data on the data base that would show problems later on when for example populating on an already deleted record.

Issue #554, opened 2022-02-24
Resolvers: Create tests for removeOrganization.js


palisadoes:
The Talawa-API code base needs to be 100% reliable. This means we need to have 100% test code coverage.

Tests need to be written for file `lib/resolvers/organization_mutations/removeOrganization.js`

- We will need the API to be refactored for all methods, classes and/or functions found in this file for testing to be correctly executed.
- When complete, all all methods, classes and/or functions in the refactored file will need to be tested. These tests must be placed in a
single file with the name `tests/resolvers/organization_mutations/removeOrganization.spec.js`. You may need to create the appropriate directory structure to do this.

### IMPORTANT:
Please refer to the parent issue on how to implement these tests correctly:

- https://github.com/PalisadoesFoundation/talawa-api/issues/490

### PR Acceptance Criteria

- When complete this file must show **100%** coverage when merged into the code base. This will be clearly visible when you submit your PR.
- [The current code coverage for the file can be found here](https://codecov.io/gh/PalisadoesFoundation/talawa-api/tree/develop/lib/resolvers/organization_mutations/). If the file isn't found in this directory, or there is a 404 error, then tests have not been created.
- The PR will show a report for the code coverage for the file you have added. You can use that as a guide.

ebrahim354:
I want to work on this please

palisadoes:
I'm going to have to revert the code again. It's causing other tests to fail.

- https://github.com/PalisadoesFoundation/talawa-api/runs/5759635905?check_suite_focus=true

Someone else is going to have to work on this.

palisadoes:
Here is the error we are getting:

![image](https://user-images.githubusercontent.com/16875803/160909538-10a96568-4271-41bd-924b-d5545b30cdef.png)

ebrahim354:
The tests are failing because they relay on the files: organizationExists, adminCheck that are changed on this PR https://github.com/PalisadoesFoundation/talawa-api/pull/653 the pull requester should fix any affected files after the refactor of the code, but I can easily fix it if you want, also I still want to work on it please.

ebrahim354:
He changed the format of errors thrown without fixing the rest of the tests that relay on it.

palisadoes:
This is getting confusing. I've reassigned the issue to you. Please resubmit the PR.

xoldyckk:
[...]

PR #654, opened 2022-03-30
Revert "resolving issue: #554"

palisadoes:
Reverts PalisadoesFoundation/talawa-api#648


palisadoes:
Here is the error we are getting:

image

PR #659, opened 2022-04-03
Fix commentsByPost

### LINKED ISSUE #658: Error: commentsByPost.js never throws an error when no comment is found
**Describe the bug**
In `lib/resolvers/post_query/commentsByPost.js`

`if (!commentFound) {
    throw new NotFoundError(
      requestContext.translate('comment.notFound'),
      'comment.notFound',
      'comment'
 );`
 
 This block of code never executes because the query returns an empty object instead of a null value.
 
 `(!commentFound)` is always false even if there was no comment found

**Expected behavior**
`!commentFound` should be true when there is no comment found

**Actual behavior**
`commentFound` contains an empty object and instead of throwing an error it skips the if block

Kunal4now:

Bugfix

**Issue Number:**

Fixes  #658 

No

**Summary**

In the current code, commentFound returns an empty object which results in `(!commentFound)` to be false everytime, even when there was no comment found

To fix this, I've changed `(!commentFound) to (!commentFound.length)`, when there is no comment found it will result in true and will throw an error.

Yes

### DIFF
--- a/lib/resolvers/post_query/commentsByPost.js
+++ b/lib/resolvers/post_query/commentsByPost.js
@@ -8,7 +8,7 @@ module.exports = async (parent, args) => {
     .populate('creator', '-password')
     .populate('post')
     .populate('likedBy');
-  if (!commentFound) {
+  if (!commentFound.length) {
     throw new NotFoundError(
       requestContext.translate('comment.notFound'),
       'comment.notFound',


palisadoes (approved):

palisadoes:
This PR needed to be reverted. The code modification caused other tests to fail. If you make a modification, please run the entire test suite.

PR #711, opened 2022-06-06
return_active_registrants from org events

guptapriyanshu7:
Only return active registrants from an event.

### DIFF
--- a/lib/resolvers/event_query/eventsByOrganization.js
+++ b/lib/resolvers/event_query/eventsByOrganization.js
@@ -58,5 +58,13 @@ module.exports = async (parent, args) => {
     .populate('tasks')
     .populate('admins', '-password');
 
+    eventResponse.forEach(
+      (event) => {
+        event.registrants = event.registrants.filter(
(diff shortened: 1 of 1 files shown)


lib/resolvers/event_query/eventsByOrganization.js
```diff
@@ -58,5 +58,11 @@ module.exports = async (parent, args) => {
     .populate('tasks')
     .populate('admins', '-password');
 
+  eventResponse.forEach((event) => {
+    event.registrants = event.registrants.filter(
+      (registrant) => registrant.status === 'ACTIVE'
```
JamaicanFriedChicken, line 63: hey @guptapriyanshu7, can you put `ACTIVE` in the constants file and call it from there.

JamaicanFriedChicken (changes requested):

lib/resolvers/event_query/eventsByOrganization.js
guptapriyanshu7: done,

PR #948, opened 2023-01-29
urgent bugFix

### LINKED ISSUE #947: Urgent, Bad Test
**Describe the bug**
npm run test when executed runs the test file for `uploadImage.ts` this file deletes the `images/` dir and can cause Data loss
**To Reproduce**

1. upload any img file from the client side
2. Everything goes great
3. run npm test for this repo
4. all the uploaded images are deleted

**Expected behavior**

**Actual behavior**

**Screenshots**

**Additional details**

kb-0311:

**Issue Number:**

Fixes #947 947

Yes
**Snapshots/Videos:**

**Summary**

**Other information**

Yes

### DIFF
--- a/tests/utilities/uploadImage.spec.ts
+++ b/tests/utilities/uploadImage.spec.ts
@@ -27,11 +27,13 @@ beforeAll(async () => {
     lastName: "lastName",
     appLanguageCode: "en",
   });
-  fs.mkdir(path.join(__dirname, "../../src/images"), (err) => {
-    if (err) {
-      throw err;
-    }
-  });
+  if (!fs.existsSync(path.join(__dirname, "../../src/images"))){
+    fs.mkdir(path.join(__dirname, "../../src/images"), (err) => {
+      if (err) {
+        throw err;
+      }
+    });
+  }
   const img =
     "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAYAAACNiR0" +
     "NAAAAKElEQVQ4jWNgYGD4Twzu6FhFFGYYNXDUwGFpIAk2E4dHDRw1cDgaCAASFOffhEIO" +
@@ -51,15 +53,6 @@ afterAll(async () => {
   fs.unlink(path.join(__dirname, "../../src/images/image.png"), (err) => {
     if (err) throw err;
   });
-  fs.rm(
-    path.join(__dirname, "../../src/images"),
-    { recursive: true },
-    (err) => {
-      if (err) {
-        throw err;
-      }
-    }
-  );
   await disconnect();
 });
 
@@ -116,6 +109,10 @@ describe("utilities -> uploadImage", () => {
       testUserObj?.image
     );
     expect(uploadImagePayload?.newImagePath).toEqual(testUserObj?.image);
+
+    fs.unlink(path.join(__dirname, "../../src/".concat(uploadImagePayload.newImagePath)), (err) => {
+      if (err) throw err;
+    });
   });
 
   it("should create a new Image when an old Image Path already Exists", async () => {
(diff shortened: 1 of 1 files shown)


noman2002:
@palisadoes @DMills27 @SiddheshKukade Please have a urgent look at this.

palisadoes (approved):

palisadoes:
This is an urgent update. We cannot afford to wait.

palisadoes:
1. @kb-0311 what was in the `src/images` directory before? Your PR didn't restore the folder, so this is really important to know.
1. Also, why are we storing uploaded images in the code tree? Is this best practice?

kb-0311:
1. Don't Worry about sensitive or crucial data being deleted , `src/image` was always in `.gitignore` so in remote, there was nothing in there to be deleted because the repo has never staged that directory.
The problem would have been if the bug had not been fixed and merged quickly, then what would have happened is after another contributor would have pulled and ran tests then all the uploaded images in the code tree would have been deleted in his local instance, which would have been a waste of time for the contributor.
2. It is definitely not the best practice however, that is the logic that is written in `uploadImage.ts` where the uploaded image is stored in code tree and  fetched via its Hashed value. That was the logic that was written when I started writing tests for it from scratch.
3. The purpose behind me deleting the dir previously in my local tree in the test file I wrote was only for development purposes- The mockfiles I was creating for testing had to be deleted as I had nothing in my local instance of `src/images` , temporarily I was deleting the images/ dir 
> Shouldn't we be testing file deletions in temporary directories
Ideally, sure but the only problem with that is how the Image uploading code is written.
https://github.com/PalisadoesFoundation/talawa-api/blob/642408e1b63388b088a6e756c01513b77ca0b10c/src/utilities/uploadImage.ts#L21-L35
To thoroughly test this file to check whether truly an image file is getting uploaded , I needed to use the `images/` dir itself or mock several utilities function. 

How was this even allowed to happen ? Because  after I committed the buggy change in the PR (which I was not aware of at that time) I was afk during which I was going through the code in my head and then I realised the bug but I thought I had time to fix it since the merge usually happens around midnight (where I am from). 

After 15 minutes , By the time I logged in on my computer I saw that the PR was already merged , very early than previous PR approval and merge timings.

I immediately went into panic mode , and so had to urgently fix inform you about the bug on slack channel, fix the bug and get it approved. Thankfully there was no damage since the fix was really swift.

I am extremely sorry.

palisadoes:
We are all learning.

Where do you think we should be storing images then? 
- Like a `static/` directory in root as I've seen in other repos? 
- or somewhere configurable in `.env`?

This would be a good bug to fix.

kb-0311:
Yes, I was already thinking about what might be the best way to tackle image uploading






I think instead of whatever we are doing now the best possible approach would be to
1. Convert the client side image to binary string and then make a mutation using that uri as string to the API.
2. In the API , we update the Object (Post ,User , Organization) image field with that binary uri .
3. whenever we need to fetch or render images we can just do it by converting the BUffer binary data string of the object image field.

This way -

1. We really reduce the size of the data because we are not sending and retrieving MBs of png files.

2. Fixes this legacy flaw that this whole fiasco revealed.

noman2002:
> Yes, I was already thinking about what might be the best way to tackle image uploading
> 
> I think instead of whatever we are doing now the best possible approach would be to
> 
>     1. Convert the client side image to binary string and then make a mutation using that uri as string to the API.
> 
>     2. In the API , we update the Object (Post ,User , Organization) image field with that binary uri .
> 
>     3. whenever we need to fetch or render images we can just do it by converting the BUffer binary data string of the object image field.
> 
> 
> This way -
> 
>     1. We really reduce the size of the data because we are not sending and retrieving MBs of png files back and forth between API and client.
> 
>     2. Fixes this legacy flaw that this whole fiasco revealed.

I think this is a better approach to tackle this problem.

PR #953, opened 2023-01-30
Use constants instead of environment variables across the app

### LINKED ISSUE #950: Use constants instead of environment variables across the app

We need to improve the code quality in the `Talawa-API` repo as we have done in `Talawa-Admin` here: https://github.com/PalisadoesFoundation/talawa-admin/pull/420

- There are many instances of  `process.env.VARIABLE_NAME` in the code.
- This is a bad coding practice since it is difficult to maintain and refactor the code

- A better approach will be to store the value extracted from the environment variable in constant files
- And the constant will be used across the app 

This seems to be the best approach

**Additional context**

- This will make it easy to update names of environment variables like done in this [commit](https://github.com/PalisadoesFoundation/talawa-admin/commit/f659baf02e55b302d41bd929523610bd37ca951b)

anshgoyalevil:
This PR introduced the use of constants instead of environment variables across the Talawa-API project.

**Issue Number:**

Fixes #950

Yes

**Snapshots/Videos:**
image

Not required.

**Summary**
This PR introduced the use of constants in place of environment variables. The constants can now be assigned the value of an environment variable, and that can be used in places where that environment variable was required.

It is not a breaking change, but yes, that is significant change.

**Other information**
No

Yes


anshgoyalevil:
@palisadoes I can write tests for the new constants added in this PR, but, that will not introduce any significant code coverage, as the those are just constant exports, which I don't think would require a new test file.

What are your suggestions sir?

noman2002 (approved):

palisadoes (approved):

PR #764, opened 2022-09-12
Backend support api ready along with AWS S3 connection.

nishnatadebnath:
**What kind of change does this PR introduce?**

**Issue Number:**

Fixes #

**Did you add tests for your changes?**

**Snapshots/Videos:**

**If relevant, did you update the documentation?**

**Summary**

**Does this PR introduce a breaking change?**

**Other information**

**Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-api/blob/master/CONTRIBUTING.md)?**

PR #956, opened 2023-01-30
init - new method of uploading image files

kb-0311:
***Work in progress***


kb-0311:
I will open this draft up for review when
1.  I have made all the necessary changes including all the files indirectly or directly affected by it .
2. Refactored and rewritten the tests for the changed code .

I am confident that I will complete the required changes for the API in 3-4 days and for the talawa mobile app and admin portal latest by next Monday.

All in all this refactoring (server + client )should be done within 7 days max.

kb-0311:
@palisadoes I was a quarter of the way done making desired changes when I made a discovery about file encoding as I was looking at methods to implement the Binary String Conversion of Image files. As it turns out my proposed solution was very naive. 

What I found was - 
1. Very roughly, the final size of Base64-encoded binary data is equal to 1.37 times the original data size . ie. My 100kb homage file when uploaded using this method was now a 137Kb string.
2. The difference of 37KB may not seem much but only when we look at from the main objective of my proposed solution - Simplicity in handling uploads and image storage in JSON format, for which the base64 encoding approach was a really popular idea in early stages of the internet when the data was not that huge.
3. But beyond that, as my solution also implied storing this string in the MongoDB document under `image` field , it would also mean I would have been essentially storing a string 137% size of any given uploaded image file which would have exponentially increased the size of our database.
4. Not only that but for the client side, the decoding <->encoding of base64 <-> Image file also causes massive overhead. Rendering and encoded base64 image is also suboptimal since encoded strings cannot be cached in the browser. Which would have significantly decreased our performance on the client side.
5. The transmission of the encoded string would have increased the bandwidth and heavily slow the request/response time.
6. Even if I compress the String using gzip , which would reduce the encoding overhead to 2.5% it still would be very large as it would still be around the size of an average file and we will be storing it in the database and querying the documents.

So I cannot continue with this approach as, unfortunately, I believe the cons of going ahead outweigh the pros . 

However, I still think the existing approach to handling image file upload is not optimal. Because essentially we are using a file system to handle image uploads.
So, well to essentially access these uploaded images , we will need to fetch them from the same server we are using to handle other services.

Not everything was a waste of time though. I learned a lot about file uploading and handling and essentially I am proposing another solution - to use Amazon s3 bucket to handle image file storage. It depends on what you think about the ide of using AWS though.

This was also being worked on https://github.com/PalisadoesFoundation/talawa-api/pull/764 here in now closed PR. May I work on this now since that seems to be a perfect implementation to solve the problem? And implement it now here under this PR?

Oh and I will be submitting a PR for the other two issues I am assigned in few hours , no point delaying them now.

kb-0311:
If you have a better idea pls feel free to add on

palisadoes:
- There is a closed PR that was dormant that used AWS for storage. Please take a look.
- We need a solution that allows the administrator to select between local storage on the API server and a remote service.
- You will need to write the code such that it will work with any future cloud storage solution. There will need to be some intermediary class and possibly plugins and/or cloud provider specific sub-classes.
- You could also consider using a hash of the file to create a unique ID that could be used to prevent/ignore uploads of files that are already in storage.
- There may also need to be a way to delete files that are not referenced by any post, user profile, etc. to reduce unwanted content on the servers.

@DMills27 @JamaicanFriedChicken @SiddheshKukade Do you have any suggestions?

palisadoes:
Here is the PR I mentioned: https://github.com/PalisadoesFoundation/talawa-api/pull/764

kb-0311:
>There is a closed PR that was dormant that used AWS for storage. Please take a look.

Yes I did check that PR out . seems it was integrating S3 in the app exclusively.
>You will need to write the code such that it will work with any future cloud storage solution. There will need to be some intermediary class and possibly plugins and/or cloud provider specific sub-classes.

Yes this seems like a great approach to work on .
I know that last year one of the projects had developed plugin architecture for the backend but I don't have any hands-on idea about plugin development , how are plugins implemented in the backend? How to go about developing plugins? any resources would be appreciated. 

>You could also consider using a hash of the file to create a unique ID that could be used to prevent/ignore uploads of files that are already in storage.

Yes there is already an imageHash model implemented for this function

xoldyckk:
@palisadoes @kb-0311 You would be introducing needless complexity by integrating s3 into the project if all you need is a way to store small images. Mongodb offers an inbuilt storage solution known as [gridFS](https://www.mongodb.com/docs/manual/core/gridfs/). It would be easier to use just this.

Though if you want better latency, high throughput etc., or the files you're storing are too large like hundreds of megabytes or even bigger then s3 is the viable choice.

PR #985, opened 2023-02-02
Separate testing database

### LINKED ISSUE #964: Feature Request: Separation of testing database

Some relevant points:-

Described above

EshaanAgg:

**Issue Number:**
Fixes #964 

**Snapshots/Videos:**

### Before Testing
before
(Only test-talawa exists, which is my primary database)

### After running `npm run test`
image
(test-talawa_TEST is automatically created and fillled with the testing data)

image
(All tests are still pass)

image
(`NODE_ENV` returns back to `undefined` as soon as the testing command is complete)

Yes.

No

**Other information**
2. No changes in GitHub Action needed as it itself uses the `npm run test` script, thus the maintainers with access to the `MONGO_DB` database can indeed verify all the data is only added to a newly created testing database.

Yes


noman2002 (commented):
Changes looks good to me. Please attach a screenshot for coverage of db.ts

EshaanAgg:
@noman2002 I don't think so `db.ts` is included in the coverage as no test file exists for it.

noman2002:
Got it.

noman2002 (dismissed):
Looks good to me.

xoldyckk:
@EshaanAgg This is too much complication for no reason. Instead of this just edit the `connect` function in `db.ts` to accept an optional `dbName` argument. Each test can then pass a `dbName` argument to the `connect` function to change database used for testing.

xoldyckk (changes requested):
@EshaanAgg This is too much complication for no reason. Instead of this just edit the connect function in db.ts to accept an optional dbName argument. Each test can then pass a dbName argument to the connect function to change database used for testing.

In mongoose.connect(uri, options) options object accepts an optional field called dbName to use a custom database instead of the default one provided in .env file. The database will be accessible on the same port under the name we provide for dbName.

EshaanAgg:
@xoldyckk The whole point of setting up the `NODE_ENV` as `testing` was to ensure that we didn't need to manually change the database for each test, and we can do it at a centralized location, in the connect function. This was the workflow that was decided with @palisadoes.

Still, if you feel that the optional argument method is better, you can let me know and I'll make the necessary changes today.

xoldyckk:
@EshaanAgg Your implementation is also error prone. You can't just append `_TEST` to the end of a connection string. Connection string comes in different formats, you can't just expect it to always end with a database name.

EshaanAgg:
I wasn't aware of this. Thanks, @xoldyckk for this pointer. Would migrate all the tests to use the optional database name parameter. 
Is `TALAWA_TESTING_DB` alright for the optional DB_NAME to be used in the tests?

xoldyckk:
> 

Yeah

xoldyckk:
> I wasn't aware of this. Thanks, @xoldyckk for this pointer. Would migrate all the tests to use the optional database name parameter. Is `TALAWA_TESTING_DB` alright for the optional DB_NAME to be used in the tests?

Yeah

palisadoes:
> > I wasn't aware of this. Thanks, @xoldyckk for this pointer. Would migrate all the tests to use the optional database name parameter. Is `TALAWA_TESTING_DB` alright for the optional DB_NAME to be used in the tests?
> 
> Here's what I suggest. Use the `NODE_ENV=testing` no problem. Check against that environment variable inside `connect` function. If `NODE_ENV === "testing"` results in `true` use the `dbName` method I mentioned.

Also, tests must only be run on the test database. Under no circumstance should tests operate on a non test database. I understand passing the db name in the function, but that may be too much flexibility for something that could potentially incapacitate a production database.

EshaanAgg:
> Here's what I suggest. Use the `NODE_ENV=testing` no problem. Check against that environment variable inside `connect` function. If `NODE_ENV === "testing"` results in `true` use the `dbName` method I mentioned.

Makes the most sense. Allows us to get the best of both worlds. Will try to implement till. This would also help to mitigate @palisadoes's concern that we must not accidentally modify the production database while testing.

EshaanAgg:
@palisadoes @xoldyckk 
Have made the required changes. Added checks to ensure that we can only connect to the custom database if and only if the environment is set to `testing`. Also updated the README.md about the same and linted files which weren't linted properly in the recent commits.

EshaanAgg:
Can the maintainers rerun the test workflow on this PR? All the tests are passing locally.

palisadoes:
You have done something to lower the overall level of code coverage. Please investigate and fix.

EshaanAgg:
@palisadoes Reverted the changes in the affected files.

palisadoes:
@xoldyckk and @noman2002 Please provide feedback and approve the PR if applicable.

noman2002 (changes requested):
Resolve the conflicting files.

EshaanAgg:
The conflicts are due to the recently merged PRs. It would be easier for me to open a new PR and modify the code there. Thus I am closing this PR here.

PR #997, opened 2023-02-04
Add documentation update for clean GitHub setup

EshaanAgg:
**What kind of change does this PR introduce?**
Feature

**Issue Number:**

Fixes #995 

**Did you add tests for your changes?**
NA

**If relevant, did you update the documentation?**
Yes

**Summary**
Add instructions to setup a clean workflow as mentioned in the linked issue.

**Does this PR introduce a breaking change?**
No

**Other information**
NA

**Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-api/blob/master/CONTRIBUTING.md)?**
Yes

PR #998, opened 2023-02-04
getSort() helper funtion

### LINKED ISSUE #994: Feature Request: Using helper functions for queries
The `getSort()` function is poorly reused where each file uses the same implementation of the function but it has been compy pasted multiple times. 

Every other query has its own getSort() function with an almost identical  implemntation.
I proposed the use of helper_function directory so that the function will have its own file and will be reused by everywhere it is defined.

create `src/reolvers/Query/helper_functions` which will contain a `getSort.ts` file whichi will export the sorting implementation used by all the queries.The files with queries which are using this getSort function will then import this helper function and then appy when it is needed. 
**Additional context**

kb-0311:

**Issue Number:**

Fixes #994 

Yes
**Snapshots/Videos:**

**Summary**

**Other information**

Yes


kb-0311:
The error is due the uploadImage.spec.ts file , a file which I had written . It is failing because the directory structure for the testing application in the github CI-CD was changed.

xoldyckk:
@kb-0311 in uploadImage.ts change
    ```
    path.join(__dirname, "../images", `/${id}-${filename}`)
    ```
to
    ```
    path.join(__dirname, "../../images", `/${id}-${filename}`)
    ```

Similarly update uploadImage.spec.ts as well.

kb-0311:
@xoldyckk why the parsing error?

palisadoes:
@kb-0311 Please fix the failing test: https://github.com/PalisadoesFoundation/talawa-api/actions/runs/4092296399/jobs/7057141172#step:7:277

This may be due to a past PR. I've noticed it happening occasionally. We now have an example of it happening and rather than re-running the GitHub action, we need to fix it.

kb-0311:
@palisadoes Hmm , Sorry but CI-CD is not something I am familiar with as of now, I rectified the testing unhandled errors and got the tests to run however there is some 'parsing error' that I do not seem to understand. Maybe someone who is more experienced with CI-CD can help us out here?

kb-0311:
it seems #997 us failing for the same reason too

kb-0311:
https://github.com/VeryGoodOpenSource/very_good_coverage/issues/77

Maybe this is helpful ?

EshaanAgg:
@kb-0311 A minor suggestion, you can consider rewriting all the branches in `getSort.ts` using `switch` statements rather than multiple if-else. They make for a better reading and debugging experience.
Another approach might be to split the `orderBy` variable on `_` and then use the first part as the property name, and the second to determine the value of 1/-1. That would make for the most concise code.

noman2002:
> it seems #997 is failing for the same reason too

it is not failing now. There is some other issue in this PR. please have look.

EshaanAgg:
> This may be due to a past PR. I've noticed it happening occasionally. We now have an example of it happening and rather than re-running the GitHub action, we need to fix it.

I can think of two ways of doing so.
1. We can change our provider for this GitHub action as we are having troubles with it. [I had previously used [this](https://github.com/barecheck/code-coverage-action) action for a project. The downside to this is obviously that these actions are way less active that our current provider.

2. In light of this [comment](https://github.com/VeryGoodOpenSource/very_good_coverage/issues/77#issuecomment-1196028216), we can add a command `[ -f ./coverage/lcov.info ] && echo "Coverage file exists" || echo "Coverage does not exist"  ` before the `VeryGoodOpenSource/very_good_coverage@v2` action to ensure that our coverage file is indeed existing before this step. If not, we can then be certain that the error is in one of the earlier testing steps, as they aren't generating the coverage file as they should.

kb-0311:
@palisadoes @noman2002 Tests are passed now, it was because the lcov file was not being properly generated earlier due to a mistake from my part how ever it is good now.

@EshaanAgg The tests are passed now , it was because of faulty vite.config change on my part. Yeah I was going to go for switch-case approach yesterday itself before my priority was to make the tests pass. I will be making the change now.

Sorry about that .

xoldyckk (commented):
You should never use `any` keyword in typescript unless you absolutely have to. Add types to `getSort` helper function parameter `orderBy`. Copy paste this in place of `any`:- 
```
| InputMaybe<
        | EventOrderByInput
        | OrganizationOrderByInput
        | PostOrderByInput
        | TaskOrderByInput
        | UserOrderByInput
      >
    | undefined
```
You also need to test the `getSort` function.

Rename `helper_funtions` folder to `helperFunctions`. Always use camelCase or PascalCase for naming stuff in javsacript/typescript code because it is the industry convention.

kb-0311:
@xoldyckk Yep added the types and naming convention you requested. 

The getSort() helper function already has 100% test coverage because it is tested by all the other queries which uses that helper.

xoldyckk (approved):
@kb-0311 Looks good.

noman2002 (approved):

PR #1016, opened 2023-02-06
Separate Testing Database [Fixes #964]

### LINKED ISSUE #964: Feature Request: Separation of testing database

EshaanAgg:

**Issue Number:**
Fixes #964 

**Snapshots/Videos:**

### Before Testing
before
(Only test-talawa exists, which is my primary database)

### After running `npm run test`
image
(test-talawa_TEST is automatically created and fillled with the testing data)

image
(All tests are still pass)

image
(`NODE_ENV` returns back to `undefined` as soon as the testing command is complete)

Yes.

No

**Other information**
2. No changes in GitHub Action needed as it itself uses the `npm run test` script, thus the maintainers with access to the `MONGO_DB` database can indeed verify all the data is only added to a newly created testing database.
3. Added checks to ensure that we can only connect to the custom database if and only if the environment is set to testing. Also updated the README.md about the same and linted files which weren't linted properly in the recent commits.
4. Re implementation of PR #985 due to merge conflicts

Yes


EshaanAgg:
@palisadoes @xoldyckk @noman2002 Can you please review this PR and merge it before merging any other PRs involving changes to testing files, as otherwise I would have to open another PR due to merge conflicts?

xoldyckk:
@EshaanAgg @noman2002 Don't you think instead of doing so many changes we could just create a seperate file to initialize mongoose?

Like make a file named `mongoose.ts` in `tests` folder, provide it with `connect`, `disconnect` functions and we explicitly set `dbName` inside that file to be `TALAWA_TEST_DATABASE`. Then we use `connect`, `disconnect` functions(or other additional functions to have more control over the test data in the database) inside tests. There won't be a need for conditional environment check inside `src/db.ts`.

EshaanAgg:
@xoldyckk That seems doable. But still we would need to change all the test files to change from where they are importing their `connect`, and `disconnect` functions? 
Then we could revert `db.ts` to as it was.

noman2002 (changes requested):
Resolve the conflicts

xoldyckk:
> 

Do it. I don't see any better way. All stuff related to setting up tests should be present within tests directory. That includes setting up/managing/shutting down the database.

Like this for example:-
```
import "dotenv/config";
import mongoose from "mongoose";

// Returns a mongoose instance
export async function connectDatabase(
  dbName: string = "TALAWA_API_TEST_DATABASE"
) {
  return await mongoose.connect(process.env.MONGODB_URI as string, {
    dbName,
  });
}

export async function dropAllCollectionsFromDatabase(
  mongooseInstance: typeof mongoose
) {
  const collections = await mongooseInstance.connection.db.collections();

  for (let collection of collections) {
    await collection.drop();
  }
}

export async function dropDatabase(mongooseInstance: typeof mongoose) {
  await mongooseInstance.connection.db.dropDatabase();
}

export async function disconnectDatabase(mongooseInstance: typeof mongoose) {
  await mongooseInstance.connection.close();
}

```

xoldyckk:
> @xoldyckk That seems doable. But still we would need to change all the test files to change from where they are importing their `connect`, and `disconnect` functions, right? Then we could revert `db.ts` to as it was.
> 
> I can get working on a new PR as soon as we agree on an approach to solve this problem.

Do it. I don't see any better way. All stuff related to setting up tests should be present within tests directory. That includes setting up/managing/shutting down the database.

Like this for example:-
```
import "dotenv/config";
import mongoose from "mongoose";

// Returns a mongoose instance
export async function connectDatabase(
  dbName: string = "TALAWA_API_TEST_DATABASE"
) {
  return await mongoose.connect(process.env.MONGODB_URI as string, {
    dbName,
  });
}

export async function dropAllCollectionsFromDatabase(
  mongooseInstance: typeof mongoose
) {
  const collections = await mongooseInstance.connection.db.collections();

  for (let collection of collections) {
    await collection.drop();
  }
}

export async function dropDatabase(mongooseInstance: typeof mongoose) {
  await mongooseInstance.connection.db.dropDatabase();
}

export async function disconnectDatabase(mongooseInstance: typeof mongoose) {
  await mongooseInstance.connection.close();
}

```

EshaanAgg:
Sure @xoldyckk. I'll configure the functions such that each test file would drop all the collections from the database once it is done running, so that the database information for the failed tests can be used to debug them.

xoldyckk:
@EshaanAgg That wasn't my intention for adding those functions. I just thought they'd be good for future use. 

Btw, currently tests are written to work against a single database and in a way that all mutation tests can be run in parallel but query tests can only work in series.

This is because each mutation test only deals with uniquely identifiable data created by it so there's never a friction between different mutation tests. 
 
Meanwhile query tests if run in parallel result in race conditions because data is mutated parallely and they don't operate on uniquely identifiable data. They have to operate on random data because that's what queries are for. So, queries tests are a good use case for cleaning database before each iteration of query test.

If there's a way to parallelize only mutation tests and keep running other tests in sequence it will greatly decrease test execution time. Getting the coverage is also to be considered.

EshaanAgg:
Got it @xoldyckk. I'll try to brainstorm a bit about all these features and logistical considerations as well, and then one a new PR implementing the required changes ASAP.

PR #1053, opened 2023-02-11
Separate Testing Database [Fixes #964]

### LINKED ISSUE #964: Feature Request: Separation of testing database

EshaanAgg:

**Issue Number:**
Fixes #964 

Yes.

No

**Other information**
2. Moved all the logic related to connecting to the testing database in the `tests` folder itself, thus ensuring that no other function connects to it.
3. Provides functions which can be later used to cleanup the database.
4. Re implementation of PR #985 due to merge conflicts, and is based on #1016 (specially [this review](https://github.com/PalisadoesFoundation/talawa-api/pull/1016#issuecomment-1422743410) by @xoldyckk)

Yes


xoldyckk (approved):

xoldyckk (changes requested):
@EshaanAgg Needs a functionality to wipe the test database before and after running all tests.

EshaanAgg:
Okay @xoldyckk! I'll try to implement the same.

EshaanAgg:
@xoldyckk Wouldn't it be better if instead of resetting the database before and after each test, instead we reset the same after each test file is run?

This has a primary advantage that if any test fails, only the data of the said test would remain in the database, and thus would be useful in debugging.

xoldyckk:
@EshaanAgg Sure. What I meant to say is that the database we're using for test should be used in a fresh state each time those tests are run. You don't want the test data piling up on the contributer's system storage. Also the more test data in database the slower tests will be because database engine has to work against useless data which doesn't relate to the current test suite in any way.

EshaanAgg:
Got it @xoldyckk. Will update the PR to do the same soon.

xoldyckk (commented):
@EshaanAgg what's the test duration before and after the change?

EshaanAgg:
I ran the tests using a local instance of MongoDB, so the duration difference was marginal.
Even in case of using Atlas, the distinction shouldn't be noticeable as we are just making ~50 (number of test suites) more calls.

xoldyckk:
@EshaanAgg I meant to ask did the test time decrease or increase compared to previous duration?

EshaanAgg:
@xoldyckk 
Before Changes: 60.21s
After Changes: 68.40s 

The time increased by approximately a factor of 10%.

palisadoes:

EshaanAgg:
@palisadoes Done.

xoldyckk:
@EshaanAgg I thought and read about this a bit more and I think the best approach is to make use of global setup/teardown scripts at least for the current way testing is done. Here's the flow:-

# global setup process
1. Connect to test database.
2. Clear previous data.
3. Run the tests in series.

# global teardown process
1. Clear the data.
2. Disconnect from database.

EshaanAgg:
Sure @xoldyckk!
Would take a deep dive into the documentation for vitest and try to implement the same.

palisadoes:
Any progress on this? It would be good to have implemented this week in anticipation of GSoC

EshaanAgg:
I am still researching the ways to implement the scripts and parallel tests that have been suggested by @xoldyckk. Sorry for the slack on my end, it's due to my university exams. However I can assure you, would resolve this issue latest by 23 Feb.

EshaanAgg:
@xoldyckk 

Here are some important updates:
1. We can't use the `globalSetup` functionality that `Vitest` provides as the global setup and teardown scripts are run in a different context than that of the test files. Thus, any mongoose connection we make in them would be dropped in the context of the test files.

2. We can't use the `setupFiles` functionality provided by `Vitest` too, as although the setup files are run in the same context as that of the test files, using the same requires the use of `beforeAll` and `afterAll` functions. Using them is not feasible for us, as we are already using `beforeAll` function in all out test files to populate the database, and using the same in the setup file to connect to the database will create a race condition.

3. Since we are resorting to connecting and disconnecting to the database for each individual test file, it does not make sense to purge data twice, in the setup as well as the teardown. Thus, I have implemented the dropping of all collections from the database only at the teardown. 

4. `Vitest` treats a single test file (say `A.spec.ts`) as a test suite, and currently only supports running all the tests in the particular test suite in a concurrent manner. It is not possible to configure `Vitest` to run multiple test files (suites) concurrently. Thus I have not implemented anything in this regard. 
If you think that we can run the particular tests in a test suite concurrently without creating a race condition, you can let me know and I will update the PR to do the same for `Mutation` resolvers.

xoldyckk:
> @xoldyckk @palisadoes
> 
> Here are some important updates:
> 
> 1. We can't use the `globalSetup` functionality that `Vitest` provides as the global setup and teardown scripts are run in a different context than that of the test files. Thus, any mongoose connection we make in them would be dropped in the context of the test files.
> 2. We can't use the `setupFiles` functionality provided by `Vitest` too, as although the setup files are run in the same context as that of the test files, using the same requires the use of `beforeAll` and `afterAll` functions. Using them is not feasible for us, as we are already using `beforeAll` function in all out test files to populate the database, and using the same in the setup file to connect to the database will create a race condition.
> 3. Since we are resorting to connecting and disconnecting to the database for each individual test file, it does not make sense to purge data twice, in the setup as well as the teardown. Thus, I have implemented the dropping of all collections from the database only at the teardown.
> 4. `Vitest` treats a single test file (say `A.spec.ts`) as a test suite, and currently only supports running all the tests in the particular test suite in a concurrent manner. It is not possible to configure `Vitest` to run multiple test files (suites) concurrently. Thus I have not implemented anything in this regard.
>    If you think that we can run the particular tests in a test suite concurrently without creating a race condition, you can let me know and I will update the PR to do the same for `Mutation` resolvers. (I don't think this would be the case, as the current tests have been designed to rely on each other)
> 5. On a local instance of MongoDB, the current testing time clocks around `64s` on an average for my laptop, and under `120s` on the PR workflow.

I agree with your point 2 and 4, but that's only because the tests within a single file(suite) are dependent on each other. If each of of them operated on their own data this wouldn't be the case. So, yeah currently we can't have them running concurrent.

About point 3, I think it makes sense to wipe db both at start and end because you can never know if the user themselves populated the test database with their own data. Also, it's not a big deal, it's not going to cost a lot of time.

EshaanAgg:
@xoldyckk 
Updated all the tests to clean the database even when it first connects to it.

EshaanAgg:
@xoldyckk @palisadoes @noman2002 
Can you please re-run the workflow and provide reviews?

xoldyckk:
@EshaanAgg I don't see the value of clearing database between mutation and relation tests. With more random data there's more assurance that mutations only mutate what they're meant to, it' not a bad thing. You're saying that it'll be easier to debug test data created in that test, it can be easily done by prefixing that data with that test's name or something similar.

I think clearing data should only be considered in query tests, to test against a pre-defined set of values which are created in that test only. This will make testing filter and sorting logic much more predictable and the test itself faster.

EshaanAgg:
@xoldyckk The approach you describe makes sense, but can we be sure in which order the tests would be running? `Vitest` "usually" runs all the tests in a directory together, but I couldn't find any point in their specification or documentation that guarantees the same. 
If we can't guarantee the same, then the purpose of not clearing the database before Mutations tests is lost, as we might be running them first.

xoldyckk:
> @xoldyckk The approach you describe makes sense, but can we be sure in which order the tests would be running? `Vitest` "usually" runs all the tests in a directory together, but I couldn't find any point in their specification or documentation that guarantees the same. If we can't guarantee the same, then the purpose of not clearing the database before Mutations tests is lost, as we might be running them first. What do you think?

HMMM, yes that's a good point. Well then there's no point in clearing data between tests. Just do it before and after all tests have run.

Many query tests currently are not constrained, like they fetch the whole table of data at once without any pagination techniques or something. That makes them slow because of data created by other tests. Until it's fixed they'll remain to be slow.

We really wouldn't need to test filtering and sorting logic because it is provided by mongoose and we don't need to test something which an external library provides. Usually the safety for such things comes from the external library providing the correct type definitions themselves(which mongoose doesn't because it sucks for typescript).

EshaanAgg:
1. Got it. I'll update the PR to not clear the database between tests, but only at the beginning and the end of all the test suites using `globalSetup` that `Vitest` provides.
2. While we shouldn't be testing the filtering and sorting logic, we must be testing that we are passing the correct arguments to our functions which are calling these inbuilt `Mongoose` sorting and filtering functions. Thus as I gather, the testing of the same becomes more of a frontend issue (obviously after we have integrated a compile-time type-checker like Prisma which uses Typescript to ensure the sanity of data flowing in our application.)

xoldyckk:
> 1. Got it. I'll update the PR to not clear the database between tests, but only at the beginning and the end of all the test suites using `globalSetup` that `Vitest` provides.
> 2. While we shouldn't be testing the filtering and sorting logic, we must be testing that we are passing the correct arguments to our functions which are calling these inbuilt `Mongoose` sorting and filtering functions. Thus as I gather, the testing of the same becomes more of a frontend issue (obviously after we have integrated a compile-time type-checker like Prisma which uses Typescript to ensure the sanity of data flowing in our application.)

Yes. We must be testing that the keywords passed as args inside resolvers map with the corresponding filter/sort object correctly. So, basically testing the getSort, getFilter functions.

EshaanAgg:
@xoldyckk 
Implemented the requested changes.

vite.config.ts
```diff
@@ -3,6 +3,7 @@ import { defineConfig } from "vitest/config";
 export default defineConfig({
   test: {
     setupFiles: ["./setupFile.ts"],
+    globalSetup: ["./tests/helpers/globalSetup.ts"],
```
xoldyckk, line 6: Leave some comments here explaining how globalSetup is working.

EshaanAgg: Done.

xoldyckk (approved):

xoldyckk:
@palisadoes You can merge now.

palisadoes (approved):

palisadoes:
@xoldyckk Why are you pushing changes for someone else's PR? This should be the responsibility of the submitter.

xoldyckk:
@palisadoes Could it result in some conflicts?

xoldyckk:
I only did it because it was a small change. I just added a few comments nothing else.

PR #1075, opened 2023-02-20
Adds non-nullable type in list and adds first and skip in connection.

### LINKED ISSUE #1052: Bug Report: Fix non-working search filters in talawa-admin
**Describe the bug**
All the filters in the organization dashboard in `talawa-admin`are non-working as of now.
We need to create specific query resolvers for

- posts `postsByOrganizationConnection` 
- events `eventsByOrganizationConnection`
- users `usersConnection`

like in the case of organizations, we created `organizationsConnection` in talawa-api and then, update the query in `talawa-admin`.

**Expected behavior**
Get the filtered instances as we try to use search filter in `talawa-admin`.

**Actual behavior**
No effect of using search filter in `talawa-admin`.

**Screenshots**

**Additional details**
Related to [talawa-admin-issue-460](https://github.com/PalisadoesFoundation/talawa-admin/issues/460)

beingnoble03:
Adds non-nullable type in list and adds first and skip in connection.

**Issue Number:**
Fixes #1052 

No, to avoid heavier PR.

No

Yes

### DIFF
--- a/src/resolvers/Query/getDonationByOrgIdConnection.ts
+++ b/src/resolvers/Query/getDonationByOrgIdConnection.ts
@@ -16,7 +16,10 @@ export const getDonationByOrgIdConnection: QueryResolvers["getDonationByOrgIdCon
     return await Donation.find({
       orgId: args.orgId,
       ...inputArg,
-    }).lean();
+    })
+      .limit(args.first!)
+      .skip(args.skip!)
+      .lean();
   };
 
 const getInputArg = (where: InputMaybe<DonationWhereInput> | undefined) => {
--- a/src/typeDefs/queries.ts
+++ b/src/typeDefs/queries.ts
@@ -30,7 +30,7 @@ export const queries = gql`
       first: Int
       skip: Int
       orderBy: EventOrderByInput
-    ): [Event]!
+    ): [Event!]!
 
     getDonationById(id: ID!): Donation!
 
@@ -41,7 +41,7 @@ export const queries = gql`
       where: DonationWhereInput
       first: Int
       skip: Int
-    ): [Donation]!
+    ): [Donation!]!
 
     getDonations: [Donation]
 
--- a/src/types/generatedGraphQLTypes.ts
+++ b/src/types/generatedGraphQLTypes.ts
@@ -1027,10 +1027,10 @@ export type Query = {
   event?: Maybe<Event>;
   events?: Maybe<Array<Maybe<Event>>>;
   eventsByOrganization?: Maybe<Array<Maybe<Event>>>;
-  eventsByOrganizationConnection: Array<Maybe<Event>>;
+  eventsByOrganizationConnection: Array<Event>;
   getDonationById: Donation;
   getDonationByOrgId?: Maybe<Array<Maybe<Donation>>>;
-  getDonationByOrgIdConnection: Array<Maybe<Donation>>;
+  getDonationByOrgIdConnection: Array<Donation>;
   getDonations?: Maybe<Array<Maybe<Donation>>>;
   getPlugins?: Maybe<Array<Maybe<Plugin>>>;
   getlanguage?: Maybe<Array<Maybe<Translation>>>;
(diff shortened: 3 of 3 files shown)


beingnoble03:
@xoldyckk @palisadoes ptal. Thanks!

xoldyckk (approved):

beingnoble03:
@xoldyckk Thanks for the resources. Should we open an issue for the correct implementation of all the connections in talawa-api?

xoldyckk:
> @xoldyckk Thanks for the resources. Should we open an issue for the correct implementation of all the connections in talawa-api?

It's not an easy task. Connections are harder to implement and you'll have to modify both admin and mobile implementations for it. But yeah they're the industry standard and way more performant when the amount of items increases.

beingnoble03:
Hmm...
Agreed. Changing both admin and mobile implementations is a tedious task but can be future goal, because bringing large amount of data in the frontend is a big problem.

palisadoes:
- How many connections would we need to modify? All?
- We have managed PR synchronization across repositories before.

xoldyckk:
> * How many connections would we need to modify? All?
> * We have managed PR synchronization across repositories before.

If the data those connections are referencing are infinetely growable then yes they'd need to be modified. Though it's a matter of scalability. When the data becomes too large database takes a heavy toll with simple pagination. That's where cursor(relay connection based) pagination shines(though it can be used for simple pagination too). If the data is not too big you can get away with using simple pagination. 

The complexity is because connections represent a different object than a plain list. The UI logic for admin and mobile is probably written considering the data they'll get will be a plain list too.

beingnoble03:
@xoldyckk Currently, in some connections in `talawa-api`, we have used `first` and `skip` as parameters instead of `first` and `cursor`. So, is this a wrong implementation of Connections?

xoldyckk:
> @xoldyckk Currently, in some connections in `talawa-api`, we have used `first` and `skip` as parameters instead of `first` and `cursor`. So, is this a wrong implementation of Connections?

Yes. They're not connections. Connection is a name specifically given to relay based cursor pagination.

xoldyckk:
> But I think there's a drawback of using the [relay based cursor pagination](https://relay.dev/graphql/connections.htm) here, as we can't jump directly to some random page number, and we have page number shown in `talawa-admin` for pagination .So, I think we need to either change it to infinite scrolling if we want to use relay based cursor pagination or continue with the offset based implementation and remove Connection from the query name. What should we do then?
> 
> I checked in the `talawa` mobile app, we have infinite scrolling there, so no issues there.

Ehh. Is having a page-based even something valuable? It's only good for non-changing data. I don't think that's true for majority of applications. You're trying to share a page to someone, but the data on that page has already been modified(moved), no point of it.

beingnoble03:
Hmm...
@palisadoes As we are planning to redesign the talawa-admin, I think we should go for infinite scrolling there, instead of page based pagination. We can start by implementing relay cursor based pagination in eventsByOrganizationConnection.. As it is only used in the talawa-admin portal. (Currently mobile app uses just one connection organizatinsConnection) And, then go for other connections.

Your views @xoldyckk @palisadoes ?

palisadoes:
There are many use cases for non-infinite scrolling, mostly in the management of resources when list views are required:

**Talawa Admin** 
- User profile management
- Organization profile management
- Calendar / event management
- Event checkins
- Tag mangement (Eventually)

**Talawa** 
- Calendar / event management
- User profile management

Is this assumption wrong, and if so why?

palisadoes:
@DMills27 @JamaicanFriedChicken @Tajcore @Kevoniat your inputs here would be welcome

DMills27:
@xoldyckk @beingnoble03  Before we consider changes to the "correct implementation for connections" throughout the codebase, we should have some heuristics in place to see if it's worthwhile. What I mean by this is how would we define a " large amount of data"? Is there some kind of load testing we can do for various components of the API to see if the difference is apparent? This would be a better next step than creating an issue to create the "proper" connectors.

Also, we need to consider if this use case can happen to us. If we end up finding out that "a large amount of data" exceeds a scale that we would realistically reach, I'm not sure there's justification in doing so.

Lastly, the load testing serves another purpose: it would show us the effect on the performance of the larger system. We need to consider how it will impact the entire system and if there are some unforeseen bottlenecks that need to be addressed.

noman2002 (approved):

xoldyckk:
> @xoldyckk @beingnoble03 Before we consider changes to the "correct implementation for connections" throughout the codebase, we should have some heuristics in place to see if it's worthwhile. What I mean by this is how would we define a " large amount of data"? Is there some kind of load testing we can do for various components of the API to see if the difference is apparent? This would be a better next step than creating an issue to create the "proper" connectors.
> 
> Also, we need to consider if this use case can happen to us. If we end up finding out that "a large amount of data" exceeds a scale that we would realistically reach, I'm not sure there's justification in doing so.
> 
> Lastly, the load testing serves another purpose: it would show us the effect on the performance of the larger system. We need to consider how it will impact the entire system and if there are some unforeseen bottlenecks that need to be addressed.

I mean yeah, it's not worth it. Unless you're scaling to thousands and millions of users there's not much performance gain to be squeezed out.

beingnoble03:
I think we can change the query names instead of keeping the wrong convention, if we are considering that page based pagination is necessary for the working of the app.

palisadoes:
The discussion has been interesting but inconclusive. Two people have approved this PR.  Should it be merged?

xoldyckk:
@beingnoble03 Change the names I guess? You'd also have to change the names from clients. It shouldn't result in any disruption.

beingnoble03:
@xoldyckk should we follow any query name convention for all the queries having `where` as a parameter (for filters)?
Removing only the "Connection" word from the query name will result in multiple queries having the same name.

xoldyckk:
@beingnoble03 Yeah I just checked out the files, this change might not be easy. There's duplicated files like `eventsByOrganization` and `eventsByOrganizationConnection`. There are similar duplicated files like these. It makes no sense to have two of these but the clients might be using them.

beingnoble03:
So, should I proceed by merging resolvers like example and exampleConnection in talawa-api? And then changing the query accordingly in the client side?

palisadoes (dismissed):
I'm requesting that @xoldyckk dismiss this request for change and merge the the PR is finally ready. That will reduce the risk of an accident as it is ready to merge as is because it has received the required number of approvals.

xoldyckk:
> I'm requesting that @xoldyckk dismiss this request for change and merge the the PR is finally ready. That will reduce the risk of an accident as it is ready to merge as is because it has received the required number of approvals.

I didn't understand what you said. I had already approved this PR 3 days ago. You can merge.

xoldyckk:
> OK, but you keep discussing stuff after, so it was unclear.

You're right. We should probably try to discuss stuff here in the future https://github.com/PalisadoesFoundation/talawa-api/discussions

PR #1081, opened 2023-02-22
Fix Regex [Fixes part of #494 in Talawa-Admin]

EshaanAgg:
Bugfix

**Issue Number:**

Fixes part of [#494 in Talawa Admin](https://github.com/PalisadoesFoundation/talawa-admin/issues/494)

NA

**Snapshots/Videos:**
NA

NA

**Summary**
Update the Regex to allow various special characters in form validation.

No

### DIFF
--- a/src/libraries/validators/validateString.ts
+++ b/src/libraries/validators/validateString.ts
@@ -1,7 +1,7 @@
 export function isValidString(str: string, maxLength: number) {
-  const pattern = /[`a-zA-Z0-9!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?~]/;
-  const isFollowingPattern: boolean = pattern.test(str);
-  const isLessThanMaxLength: boolean = str.length < maxLength;
+  const pattern = /[a-zA-Z0-9!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?]/;
+  const isFollowingPattern: boolean = pattern.test(str) && str.length !== 0;
+  const isLessThanMaxLength: boolean = str.length <= maxLength;
   return {
     isFollowingPattern,
     isLessThanMaxLength,
--- a/src/libraries/validators/validateString.ts
+++ b/src/libraries/validators/validateString.ts
@@ -1,5 +1,5 @@
 export function isValidString(str: string, maxLength: number) {
-  const pattern = /^[a-zA-Z0-9!@#$%^&*()_\-+. ,]+$/;
+  const pattern = /[`a-zA-Z0-9!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?~]/;
   const isFollowingPattern: boolean = pattern.test(str);
   const isLessThanMaxLength: boolean = str.length < maxLength;
   return {
--- a/tests/libraries/validators/validateString.spec.ts
+++ b/tests/libraries/validators/validateString.spec.ts
@@ -1,0 +1,29 @@
+import { isValidString } from "../../../src/libraries/validators/validateString";
+import { describe, it, expect } from "vitest";
+
+describe("Validators -> isValidString", () => {
+  it("Check the max length response", () => {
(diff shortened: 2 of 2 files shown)


palisadoes:
Please create or update the tests for this new validation regex

EshaanAgg:
Done @palisadoes. Please review.

palisadoes:
What specifically is this validating? 

1. Usernames
1. Email addresses
1. Passwords
1. Other

It appears we'll need separate ones for each type based on the original issue.accordingly.

xoldyckk:
@EshaanAgg This approach of using regexes to validate inputs is really dangerous. Regexes should be avoided almost always and even if they're being used, they should be hidden behind an abstraction like a validator library which has good reputation. They're not bad, just that they're really hard to maintain and usually people will write bad regexes which will hurt app performance. 

Writing custom validators for each data type is time consuming. This also doesn't provide an easy way to create/format errors to send to the client. Instead each field accepting arguments should have its own argument parser function which does all the hard work and gives back either validated/formatted data or a list of errors which can be sent to client.

There should be some constants like minimum/maximum values for Int types, minimum/maximum values for String types.

Also, error object's shape for args validation should be consistent/immutable and provide proper details to the client about errors so, that clients can implement UI logic according to that.

EshaanAgg:
@palisadoes Currently, the same validator is being used across all the string input types, and thus we can't make the Regex more specific.

@xoldyckk I agree that we should ditch the concept of using Regex in entirety. I proposed that we move forward with the integration of [validator](https://www.npmjs.com/package/validator) into our backend, which is a strictly strings only library with 100% code testing.
As for this issue, I can add this library to our package, and then work on establishing a scable flow for validating the arguments that we receive in the `createOrganizarion` mutation and providing the front-end with user-friendly error messages. 
Then after review, we can migrate to using the same approach for all mutations (basically anytime we are writing or updating our database).

palisadoes:
Yes regexes are a pain.

xoldyckk:
> @palisadoes Currently, the same validator is being used across all the string input types, and thus we can't make the Regex more specific.
> 
> @xoldyckk I agree that we should ditch the concept of using Regex in entirety. I proposed that we move forward with the integration of [validator](https://www.npmjs.com/package/validator) into our backend, which is a strictly strings only library with 100% code testing. I don't think so we need any strict validations with numbers as of now, and would let you know if I find any such instance during the updation.
> 
> As for this issue, I can add this library to our package, and then work on establishing a scalable flow for validating the arguments that we receive in the `createOrganizarion` mutation and providing the front-end with user-friendly error messages. Then after review, we can migrate to using the same approach for all mutations (basically anytime we are writing or updating our database).

Basically if you want more flexible control over the shape and customization of your error objects you would use scalar validators and use their parsed results to determine what errors should be sent to the client and nicely shape them up. A validation library like zod or yup would return their own shape of errors and it's tedious to format it to our required shape.

Btw you should check out graphql-scalars. It provides many common scalars to inject into graphql schema, so you probably won't even need a scalar validation library.

EshaanAgg:
Ohh wow. This seems really powerful!

But still, this is not a complete solution, as scalars are sort of type definitions for our schema, but we still need validators. A simple example of the same is the description field. It can be simply implemented with a `string` scalar, but we still need to add some validator on it (to ensure the max length, min length, and other relevant properties). Other validators, like uniqueness validators for tags and other needs to be implemented by us, or a third-party library.

I feel that we should try to adopt GraphQL scalars into our schema, as they make the same more robust and allow us to leverage the complete power of the ecosystem that has been developed by `The Guild`, but the same should not be treated as a validation library, which is necessary to any backend repository. The inclusion of the same though would vastly help us with the parsing and proper formatting of the data.

xoldyckk:
@EshaanAgg Yes, that's what I said we don't need to validate the data type in most cases as that would be done by graphql-scalars. We need to validate business requirements.

Issue #1082, opened 2023-02-23
Implement a Scalable workflow for validation in Mutations


EshaanAgg:
**Is your feature request related to a problem? Please describe.**
As evident from the discussion [here](https://github.com/PalisadoesFoundation/talawa-api/pull/1081), there is a need for:
1. Add GraphQL scalars to talawa-api repository so that most of the data types can be validated for us.
2. Add validation methods to check for business requirements in all of the mutation methods, and provide a way to easily send error messages for the same. 

**Describe the solution you'd like**
Integrating GraphQL scalars into the repository seems like a job that can been accomplished within a short time frame, and thus should be done ASAP. After the integration of the same, we must come to consensus about a uniform strategy for validation (a third party library maybe?), and then try to incorporate the same in the mutation resolvers (as well as add tests for them).
 
**Approach to be followed (optional)**
Described above

**Additional context**
I can get started on the GraphQL scalars integration, and in the meanwhile we can try to ideate the best way to deal with validation requirements.

palisadoes:
Are you still working on this? It's been over 10 days. Others need to get a chance.

EshaanAgg:
This issue has already been implemented by other contributors in other PRs with the help of regexes.
As per discussions with @xoldyckk, we should move away from regexes and do so with the help of some validation library, but it would make sense to implement it only after the migration to `GraphQL Yoga`.

So as of now, this issue can be closed.

PR #1085, opened 2023-02-23
Add GraphQL-Scalars [Fixes part of #1082]

EshaanAgg:
Add the package GraphQL scalars to the codebase, and migrates the relevant fields on the Schema to use the same

**Issue Number:**

Fixes part of #1082 
The part of validation would be fixed in another PR, so that it becomes easy to review the same.

NA

NA

**Summary**
Provides easy method of parsing and working with some fields.

YES

**Other information**
- We can't simply use the `scalarTypeDefs` as described in the official documentation [here](https://the-guild.dev/graphql/scalars/docs/usage/apollo-server) as then we recieve a typescript error (our `typeDefs` are of the type `DocumentNode` while those of `graphql-scalars` are of `string`. Thus to combat the same, I have defined the scalars manually in `scalars.ts` and then configured the `graphql-scalar` resolvers for the same in `resolvers/index.ts`)

Yes


EshaanAgg:
Working on this PR, I have also noticed that we have defined the scalar `Upload` in our schema, but I couldn't find any resolver for this custom scalar anywhere in our codebase, even when this scalar is being used in the Schema definition for `file` type.

We must change the same after an approach to deal with `Image` utilities is finalised.

xoldyckk:
@EshaanAgg Read https://www.apollographql.com/blog/graphql/file-uploads/with-apollo-server-2-0/

Since, `scalar Upload` is only added by apollo server at runtime, graphql code generator cannot possibly know that because it looks in typeDefs folder for the scalar while generating types. It is added so that graphql code generator doesn't throw error when generating types for it.

I guess it shouldn't exist in typeDefs. Instead it should exist in graphql codegen config like this:
```
 schema: ["./src/typeDefs/**/*.ts", "scalar Upload"]
```

The typescript type generated by graphql-codegen for this scalar will be `any`.

EshaanAgg:
Got it @xoldyckk! I'll move `scalar Upload`  from the `typeDef` file to `codegen` as suggested by you, as it makes much more sense to keep it seperate as it is added at runtime. 

> Btw that error is your local problem as it doesn't happen for anyone else while running `npm run dev`.
Oh wow. That's actually great lol. I will reset my local development environment and try to figure out why it was happening. Meanwhile you can review the PR and let me know if you find any scope for improvements.

src/typeDefs/inputs.ts
```diff
@@ -325,7 +325,7 @@ export const inputs = gql`
   input PostUpdateInput {
     text: String
     title: String
-    imageUrl: String
-    videoUrl: String
+    imageUrl: URL
```
xoldyckk, line 328: This implementation is probably flawed btw whereever it's used. If the client wants to update their post they need to send back a new image or video, not a url. There are many other places too where this way of updating objects is used.

src/resolvers/index.ts
```diff
@@ -19,4 +30,15 @@ export const resolvers: Resolvers = {
   Organization,
   Query,
   Subscription,
+
+  // graphql-scalar resolver
+  Date: DateResolver,
+  DateTime: DateTimeResolver,
```
xoldyckk, line 36: Please try to place items lexographically wherever possible. It's makes it easier to navigate and takes less mental overhead to make sense of things.

src/typeDefs/types.ts
```diff
@@ -29,7 +29,7 @@ export const types = gql`
   type Comment {
     _id: ID
     text: String!
-    createdAt: String
+    createdAt: Timestamp
```
xoldyckk, line 32: Don't use timestamp. Replace all instances with DateTime.

xoldyckk (commented):
I'd also like to remind you that this could break some client implementations if they are sending improper/dummy data to the api while testing or just trying things out. You'll have to make sure they don't break.

src/typeDefs/types.ts
EshaanAgg: Okay. Any specific reason for this though?

src/resolvers/index.ts
EshaanAgg: All the items are placed loexographically already. I wanted to keep the graphql-scalar resolvers and our imported resolvers separate, that's why I didn't merge them. 

Should I mix both of them, and then sort them lexographically?

src/typeDefs/inputs.ts
EshaanAgg: I was thinking the same. But currently I do not think this implementation is functional, both in admin as well as in talawa. 
So I think its better we leave it as a URL, till we can standardise their implementations across the projects.

src/typeDefs/types.ts
xoldyckk: @EshaanAgg They can't be used for localised date. You can't change them dynamically based on client's time zone and stuff.

src/resolvers/index.ts
xoldyckk: Hmm, let it be no problem.

src/typeDefs/inputs.ts
xoldyckk: Ok

src/typeDefs/types.ts
EshaanAgg: Ohh, wow. Wasn't aware about the same. Made the required changes in the latest commit.

xoldyckk (dismissed):

EshaanAgg:
> I'd also like to remind you that this could break some client implementations if they are sending improper/dummy data to the api while testing or just trying things out. You'll either have to make sure they don't break or someone working on the client side has to fix their code.

I'm completely aware about the same. But this migration was long time pending. Hopefully once this is published, we can work on the client side to find possible errors and implement quick bug fixes to resolve the same. Do you think there is any way we can start testing the same in a structured way?

xoldyckk:
> > I'd also like to remind you that this could break some client implementations if they are sending improper/dummy data to the api while testing or just trying things out. You'll either have to make sure they don't break or someone working on the client side has to fix their code.
> 
> I'm completely aware about the same. But this migration was long time pending. Hopefully once this is published, we can work on the client side to find possible errors and implement quick bug fixes to resolve the same. Do you think there is any way we can start testing the same in a structured way?

Testing what?

EshaanAgg:
The breaking of the client side applications due to these changes in the schema.

EshaanAgg:
Update: Found a breaking change
image
Upon trying to register a new user (or even login) from admin panel, the API now logs this error while the client side application shows an error of this form.
image

How should we go about fixing this and similar errors that might possibly arise in the future?

xoldyckk:
> The breaking of the client side applications due to these changes in the schema.

Hehe, there's not way to test it except for manual testing I guess? Unless the clients write their main code and test code correctly there's nothing to be done.

xoldyckk:
@palisadoes do not merge yet. It will break many things on the client side.

palisadoes:
@xoldyckk Change your approval to "request changes" and dismiss them when ready. That really helps the process

EshaanAgg:
I have added the relevant changes in the Admin repository to make this PR compatible with the same (tested the same locally, and the functionality now works smoothly)

The same changes need to be made for the mobile app too (?). The only changes are needed in the Scalar definition of the arguments of the mutations that are being called, as shown in the above PR. Having no experience in the Mobile development, the talawa repo is a little hard for me to navigate. If anyone working on the same can help me, I can make the changes for compatibility in it too.

xoldyckk:
> I have added the relevant changes in the Admin repository to make this PR compatible with the same (tested the same locally, and the functionality now works smoothly)
> 
> The same changes need to be made for the mobile app too (?). The only changes are needed in the Scalar definition of the arguments of the mutations that are being called, as shown in the above PR. Having no experience in the Mobile development, the talawa repo is a little hard for me to navigate. If anyone working on the same can help me, I can make the changes for compatibility in it too.

You can create an issue on talawa repo and link this PR as well. Someone will surely pick it up.

> I have added the relevant changes in the Admin repository to make this PR compatible with the same (tested the same locally, and the functionality now works smoothly)
> 
> The same changes need to be made for the mobile app too (?). The only changes are needed in the Scalar definition of the arguments of the mutations that are being called, as shown in the above PR. Having no experience in the Mobile development, the talawa repo is a little hard for me to navigate. If anyone working on the same can help me, I can make the changes for compatibility in it too.

I'll make an issue on talawa mobile repo.

EshaanAgg:
Thanks @xoldyckk!

xoldyckk:
@EshaanAgg The current mongoose implementation uses Timestamps instead of DateTime. I'd say change that too.

EshaanAgg:
> @EshaanAgg The current mongoose implementation uses Timestamps instead of DateTime. I'd say change that too.

@xoldyckk I can't seem to find any such instance. Currently, fields like `createdAt` are defined to be of the `scalar DateTime` and their default value is set to be `Date.now`. This results in them being stored in `ISO 8601` format in the database as shown.
image

Did I understood you correctly, or there is some other implementation that you would like me to change?

xoldyckk:
> > @EshaanAgg The current mongoose implementation uses Timestamps instead of DateTime. I'd say change that too.
> 
> @xoldyckk I can't seem to find any such instance. Currently, fields like `createdAt` are defined to be of the `scalar DateTime` and their default value is set to be `Date.now`. This results in them being stored in `ISO 8601` format in the database as shown. image
> 
> Did I understood you correctly, or there is some other implementation that you would like me to change?



> > @EshaanAgg The current mongoose implementation uses Timestamps instead of DateTime. I'd say change that too.
> 
> @xoldyckk I can't seem to find any such instance. Currently, fields like `createdAt` are defined to be of the `scalar DateTime` and their default value is set to be `Date.now`. This results in them being stored in `ISO 8601` format in the database as shown. image
> 
> Did I understood you correctly, or there is some other implementation that you would like me to change?

That's UTC time zone. The standard default time zone. So, it's the same thing as storing a timestamp and just converting it to datetime with UTC time zone. The local time zone information is lost.

So, on the client side you'd calculate the datetime relative to UTC not relative to local time zone it was created in. That'd be incorrect relative time.

EshaanAgg:
Got it @xoldyckk.

EshaanAgg:
@xoldyckk I have added the package `dayjs` to this project as well, as it is the same package which is being used by `talawa-admin` to manage the date objects. 
I have also updated the default values of all `createdAt` attributes to now first parse the `Date.now` method with this library, so that the dates are stored in the database in the ISO format, but after being converted to the UTC timezone, so that when they are parsed on the front-end, the relative time is correct.

Eg: I created at organisation at 15.32 in India (+5.30 UTC), and it was converted to the relevant UTC timezone before being stored in the database.
image

xoldyckk:
> @xoldyckk I have added the package `dayjs` to this project as well, as it is the same package which is being used by `talawa-admin` to manage the date objects. I have also updated the default values of all `createdAt` attributes to now first parse the `Date.now` method with this library, so that the dates are stored in the database in the ISO format, but after being converted to the UTC timezone, so that when they are parsed on the front-end, the relative time is correct.
> 
> Eg: I created at organisation at 15.32 in India (+5.30 UTC), and it was converted to the relevant UTC timezone before being stored in the database. image

@EshaanAgg Hmm, this looks problematic. If we really want to store datetime with local time zone we'd need the clients to send their own datetime during object creation. You can't generate datetime with time zone information on the server because server is not the source of object creation here. For createdAt fields it doesn't make sense I guess.

Also, mongodb internally stores all dates as unix epoch milliseconds so that time zone information is lost anyway. If you want it to persist you'd have to either store the date as string or make a seperate field for storing time zone explicitly. Too much complexity not much to gain.

Revert the mongoose model changes.

Datetime should only be used for datetime fields which are explicitly provided by the clients. For example the `deadline` argument.

EshaanAgg:
Got it @xoldyckk. Reverted the changes.

palisadoes:
- As much as possible we should be using UTC to track time related data in the DB. Make the client apps do the conversion to the local time.
- There are many metropolitan areas in the USA that span two time zones. It could make the apps unusable.

EshaanAgg:
@palisadoes @xoldyckk I have tested the time zone functionality and arrived at the conclusion we need no other changes than the current proposed ones in the current PR's. 

The date attributes like `createdAt` are being generated on the backend itself, and thus are always converted to UTC before being stored in DB. On the client side, we are parsing the same with `dayjs` library, which itself converts the time to the local time zone of the user. Thus the current functionality is well equipped to handle different time zones.

palisadoes (approved):

xoldyckk:
> @palisadoes @xoldyckk I have tested the time zone functionality and arrived at the conclusion we need no other changes than the current proposed ones in the current PR's.
> 
> The date attributes like `createdAt` are being generated on the backend itself, and thus are always converted to UTC before being stored in DB. On the client side, we are parsing the same with `dayjs` library, which itself converts the time to the local time zone of the user. Thus the current functionality is well equipped to handle different time zones.

For mongodb object ids:- https://the-guild.dev/graphql/scalars/docs/scalars/object-id

This way business logic(min/max id length) for validating ids won't be required. 

This won't bind us to only objectids because we also have other options like Int, UUID and CUID on graphql-scalars. So, we can swap objectid with these values later if we want.

EshaanAgg:
Got you. I'll update both the client and backend to use the same.

xoldyckk (commented):
Yep. It's good.

palisadoes:
@xoldyckk please approve if necessary

xoldyckk:
@EshaanAgg I just read about using ids in graphql and it seems that the standard is using base64 encoded ids. So, that it remains opaque to clients what ids represent. We'll have to validate it manually.

There's also the concept of base64 encoding the ids with __typename of the objects that implement the node interface. Node interface implementation makes it so that all the objects in graphql schema that implement node interface are queriable from a single query field called node. 

When we base64 encode an object's id with its __typename it becomes a globalId. Any globalId on the client can be used to fetch the objects that implement node interface.

Looks like this:-
```
User:12345                 ->  VXNlcjoxMjM0NQ==
```

We encode user's database id 12345 with its __typename. 

Similarly for organization:-
```
Organization:21313  ->  T3JnYW5pemF0aW9uOjIxMzEz 
```
These are global ids and they all can be used to query their respective db objects using a single query field.

EshaanAgg:
@xoldyckk 
Wow! The notion of global IDs is so fascinating!
Though reading the docs I have a couple of questions:

1. Do you think we should implement a new scalar for storing these `base64` hashed global IDs, or just using a string is enough?

2. Would we shifting all our IDs to using these hashed global ids (like for querying the Org:12345 in a getOrganisationById for example, would we use the id = 12345, or its equivalent base64 id)?

3. Can you provide me a bit more clarity as to what the resolver for the node type might look like? 
Would it look something like we would use the __typename as a switch to map to the database table, and then return the complete object with the id in that table?

xoldyckk:
> @xoldyckk Wow! The notion of global IDs is so fascinating! Though reading the docs I have a couple of questions:
> 
> 1. Do you think we should implement a new scalar for storing these `base64` hashed global IDs, or just using a string is enough?
> 2. Would we shifting all our IDs to using these hashed global ids (like for querying the Org:12345 in a getOrganisationById for example, would we use the id = 12345, or its equivalent base64 id)?
> 3. Can you provide me a bit more clarity as to what the resolver for the node type might look like?
>    Would it look something like we would use the __typename as a switch to map to the database table, and then return the complete object with the id in that table?
> 4. Do we plan on using relay on our client side apps in the near future?
> 5. Seeing this PR to be approved, should I shift this discussion to a new issue?

Base64 ids are just ASCII strings, but we really shouldn't validate that they're base64 because it's anti pattern. Unique ids should be annotated with `ID` only. 

They'll be validated manually in resolvers.

We'll decode the ids manually, seperate the __typename and id by `:`. Check if __typename is a valid node type and also whether id is a valid. This is where we can check whether id is a valid objectId(or uuid, cuid etc).

xoldyckk:
> @xoldyckk Wow! The notion of global IDs is so fascinating! Though reading the docs I have a couple of questions:
> 
> 1. Do you think we should implement a new scalar for storing these `base64` hashed global IDs, or just using a string is enough?
> 2. Would we shifting all our IDs to using these hashed global ids (like for querying the Org:12345 in a getOrganisationById for example, would we use the id = 12345, or its equivalent base64 id)?
> 3. Can you provide me a bit more clarity as to what the resolver for the node type might look like?
>    Would it look something like we would use the __typename as a switch to map to the database table, and then return the complete object with the id in that table?
> 4. Do we plan on using relay on our client side apps in the near future?
> 5. Seeing this PR to be approved, should I shift this discussion to a new issue?

Yes, we decode the id, check the __typename and fetch the database object that refers to it. Whether we plan on using relay or not doesn't matter. What matters is following best practices.

xoldyckk:
@palisadoes Why did you merge this? This will break talawa mobile probably.

noman2002:
> @palisadoes Why did you merge this? This will break talawa mobile probably.

Its not merged. Its closed.

xoldyckk:
> > @palisadoes Why did you merge this? This will break talawa mobile probably.
> 
> Its not merged. Its closed.

Huh. Merged it on client and closed it on server. Why?

palisadoes:
I'm not sure why this happened. I was merging other PRs with my phone, not closing them.

EshaanAgg:
@xoldyckk
I think we should take this new feature of being able to query by global IDs to a new issue, so that this PR can be merged, as other contributors have already started working with the latest pull from Talawa-Admin, and thus are facing breaking changes due to this PR not being merged in the API.

xoldyckk:
> @xoldyckk I think we should take this new feature of being able to query by global IDs to a new issue, so that this PR can be merged, as other contributors have already started working with the latest pull from Talawa-Admin, and thus are facing breaking changes due to this PR not being merged in the API.

What about talawa mobile?

EshaanAgg:
I contacted @literaleval, and he would be making the corresponding changes for the same and making a PR on priority basis today itself.

xoldyckk:
> I contacted @literalEval, and he would be making the corresponding changes for the same and making a PR on priority basis today itself.

That `objectId` thing needs to go away. Revert it here and on admin as well.

EshaanAgg:
@xoldyckk Done. Made the relevant changes in this PR, and added a new one for Talawa-Admin.

xoldyckk:
> I contacted @literalEval, and he would be making the corresponding changes for the same and making a PR on priority basis today itself.

So, should this be merged right now?

EshaanAgg:
I guess yes. It would make talwa-admin functional again.

xoldyckk (approved):

palisadoes (approved):

EshaanAgg:
> > @xoldyckk Wow! The notion of global IDs is so fascinating! Though reading the docs I have a couple of questions:
> > 
> > 1. Do you think we should implement a new scalar for storing these `base64` hashed global IDs, or just using a string is enough?
> > 2. Would we shifting all our IDs to using these hashed global ids (like for querying the Org:12345 in a getOrganisationById for example, would we use the id = 12345, or its equivalent base64 id)?
> > 3. Can you provide me a bit more clarity as to what the resolver for the node type might look like?
> >    Would it look something like we would use the __typename as a switch to map to the database table, and then return the complete object with the id in that table?
> > 4. Do we plan on using relay on our client side apps in the near future?
> > 5. Seeing this PR to be approved, should I shift this discussion to a new issue?
> 
> Yes, we decode the id, check the __typename and fetch the database object that refers to it. Whether we plan on using relay or not doesn't matter. What matters is following best practices.
> 
> This implementation makes the graphql schema graphql client agnostic. Any client can be used to its full potential with this.
@xoldyckk 
How do you think the implementation of the same would look like?

As a first easy step, I think so we would have to create a field level resolver on the `id` property for each of our `types`, and that resolver would basically just `base64` encode the string `__typename:MONGO_ID` and return the same. Implementing the node interface would also be an easy job, as we would just have to create a resolver that `base64` decodes the sent `id` and then returns the correct object on the basis of `__typename` and `MONGO_ID`.

It is the queries and mutations that we have implemented that concern me. To implement them, we would have to change every single query and mutation that accepts `id` as an argument to first `base64` decode the hashedID on which now our system operates, then check the `__typename` to match the supposed type it should have, and then treat the retrieved ID as the new id for the subsequent steps.

While we can implement the same, it will certainly more our code longer. Also the code would lead to duplication, as each query/mutation accepting ID would be doing the same process which is described above. 

Does this cost justify the gain of `graphql client agnostic server`? Also do you think there is a better way to do the cycle described above for all the queries/mutations?

xoldyckk:
> > > @xoldyckk Wow! The notion of global IDs is so fascinating! Though reading the docs I have a couple of questions:
> > > 
> > > 1. Do you think we should implement a new scalar for storing these `base64` hashed global IDs, or just using a string is enough?
> > > 2. Would we shifting all our IDs to using these hashed global ids (like for querying the Org:12345 in a getOrganisationById for example, would we use the id = 12345, or its equivalent base64 id)?
> > > 3. Can you provide me a bit more clarity as to what the resolver for the node type might look like?
> > >    Would it look something like we would use the __typename as a switch to map to the database table, and then return the complete object with the id in that table?
> > > 4. Do we plan on using relay on our client side apps in the near future?
> > > 5. Seeing this PR to be approved, should I shift this discussion to a new issue?
> > 
> > 
> > Yes, we decode the id, check the __typename and fetch the database object that refers to it. Whether we plan on using relay or not doesn't matter. What matters is following best practices.
> > This implementation makes the graphql schema graphql client agnostic. Any client can be used to its full potential with this.
> 
> @xoldyckk How do you think the implementation of the same would look like?
> 
> As a first easy step, I think so we would have to create a field level resolver on the `id` property for each of our `types`, and that resolver would basically just `base64` encode the string `__typename:MONGO_ID` and return the same. Implementing the node interface would also be an easy job, as we would just have to create a resolver that `base64` decodes the sent `id` and then returns the correct object on the basis of `__typename` and `MONGO_ID`.
> 
> It is the queries and mutations that we have implemented that concern me. To implement them, we would have to change every single query and mutation that accepts `id` as an argument to first `base64` decode the hashedID on which now our system operates, then check the `__typename` to match the supposed type it should have, and then treat the retrieved ID as the new id for the subsequent steps.
> 
> While we can implement the same, it will certainly more our code longer. Also the code would lead to duplication, as each query/mutation accepting ID would be doing the same process which is described above.
> 
> Does this cost justify the gain of `graphql client agnostic server`? Also do you think there is a better way to do the cycle described above for all the queries/mutations?

Yes, the implementation would look like what you described, we'll have to encode/decode it for most query, mutation and relation fields. It will be redundant.

GraphQL isn't something too wonderful. It just shifts most of the complexity from client side to server side. Clients only need to care about fetching the data however they want. They needn't care about different rest endpoints, making seperate requests to different endpoints to fetch different pieces of data for a particular page. All they have to do is make a single query with all the data they need for a single page, pass the data fetched exactly in the shape the child components requested it using fragments and be done with it. How the data they requested is resolved on the server isn't their concern. 

If we don't use globalIds there's no node interface, then we'll have to create seperate query fields for each object like this:-
```
type Query {
  user (id: ID!): User
  organization (id: ID!): Organization
  event (id: ID!): Event
  post (id: ID!): Post
  comment (id: ID!): Comment
}
```
How they'll manage it on client side:-
```
query User($id: ID!){
  user (id: $id){
  }
}

query Organization($id: ID!){
  organization (id: $id){
  }
}

query Event($id: ID!){
  event (id: $id){
  }
}

query Post($id: ID!){
  post (id: $id){
  }
}

query Comment($id: ID!){
  comment (id: $id){
  }
}
```
This shifts the complexity of managing different query fields on the client.

With node interface it's as easy as this:-
```
type Query {
  node(id: ID!): Node
}
```
Now how they'll fetch it on the client side:-
```
query Node($id: ID!) {
  ...on User {
  }
  ...on Organization {  
  }
   ...on Event {  
  }
  ...on Post { 
  }
  ...on Comment {
  }
}
```

You see using node interface is a good pattern for data fetching on the client and it is compatible with any client side data fetching library out there. 

Relay graphql client takes this pattern a bit further. While in other graphql clients you'll have to manually implement the refetch logic for stale/mutated data(whether you use node interface or not). If you provide it a node interface query field to relay you don't have to implement that logic. This is because relay has a special meaning for node interface.  It'll automatically detect which fields in your graphql schema implement node interface and automatically provide you the functionality for refetching stale/mutated data for that object which implements node interface.

EshaanAgg:
No doubt that the node interface is a very powerful tool that has been made a part of the `GraphQL` specification, but even I feel that the true power of the same can only be leveraged when we update our client apps to use the same too. 
But never the less, we should go forward with implementing these global IDs and providing support for them in our queries and mutations. 

Should I open a new issue for the same?

xoldyckk:
> No doubt that the node interface is a very powerful tool that has been made a part of the `GraphQL` specification, but even I feel that the true power of the same can only be leveraged when we update our client apps to use the same too. But never the less, we should go forward with implementing these global IDs and providing support for them in our queries and mutations.
> 
> Should I open a new issue for the same?

Not yet, I don't think it's beneficial right now. Our first focus should be getting rid of outdated/deprecated packages, and follow better typescript standards.

EshaanAgg:
True that. This could be something of a future prospect hopefully!

PR #1098, opened 2023-02-28
Add ability to toggle posts as pinned and unpinned [Fixes #1096]

### LINKED ISSUE #1096: Create the ability to toggle posts as pinned or unpinned

1. The Talawa home screen shows a newsfeed of  posts from Talawa users.
1. In the Talawa Admin console there are plans to have the ability for administrators to not only add posts to the news feed, but to also pin them to the top of the newsfeed
1. In anticipation of that feature we need the API to support this requirement
1. Look at the video below to see how the pinned posts are expected to be rendered on the Talawa mobile app screen

1. Create ways to:
    1. store the pinned characteristic to every post
    1. update this characteristic to have one of two states, pinned or unpinned
    1. determine whether a post is pinned or not for the mobile app.
    1. create a way for the mobile app to get the N th. currently pinned post 
        1. always in reverse chronological order
        1. indicating whether the post is the last one in the list
1. 100% test code coverage 
1.  Update these Talawa-Docs  as necessary (This will require another issue created in that repo by you)
     1. [Schemas Page](https://docs.talawa.io/docs/developers/talawa-api/schemas) 
     1. [Mutations Page](https://docs.talawa.io/docs/developers/talawa-api/mutations) 
     1. [The markdown versions of those files can be found here](https://github.com/PalisadoesFoundation/talawa-docs/tree/develop/docs/developers/talawa-api). 

N/A

See above

**Additional context**
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=ix3eSReJj2k?t=79)

EshaanAgg:
Adds feature to implement marking posts as pinned and unpinned 

**Issue Number:**

Fixes #1096 

Yes

**Snapshots/Videos:**
NA

Will update the documentation after I get a sanity check for the mutations and queries implemented.

**Summary**
The PR introduces the following changes:
1. Add a `pinned` field on each post which is a boolean
2. Add two mutations, `pinPost` and `unpinPost`, each of which take in the `id` of the post as an argument and then sets the status of `pinned` property of the post to be `true` and `false` respectively.
3. Implemented the `pinnedPosts` query which returns all the posts which are pinned, latest first.
4. Implemented the `pinnedPostsByOrganization`  query, which returns all the pinned posts for a particular orginzation, latest first.

NO

**Other information**
Currently we have a `postsByOrganizationConnection` query defined in our schema too. Do we need a corresponding query for `pinnedPosts` too?

Yes


noman2002 (approved):
Looks good to me.

Kevoniat:
looks good to me

xoldyckk:
@palisadoes Don't merge yet. Wait for the discussion to conclude.
https://github.com/PalisadoesFoundation/talawa-api/issues/1096#issuecomment-1448421491

xoldyckk (commented):
@EshaanAgg Take a look at this discussion. Some things can be improved.
https://github.com/PalisadoesFoundation/talawa-api/issues/1096

xoldyckk:
@EshaanAgg You can make the changes as discussed. Also, just include pagination for pinnedPosts by default, it's no big deal.

EshaanAgg:
Got it @xoldyckk.

src/resolvers/Mutation/createPost.ts
```diff
@@ -93,6 +94,38 @@ export const createPost: MutationResolvers["createPost"] = async (
     imageUrl: args.file ? uploadImageObj?.newImagePath : "",
   });
 
+  // Check if the pinned property is set
+  if (args.data.pinned) {
```
xoldyckk, line 98: Make this `args.data.pinned === true`. We don't need to check all other cases.

src/resolvers/Mutation/createPost.ts
```diff
@@ -93,6 +94,38 @@ export const createPost: MutationResolvers["createPost"] = async (
...
+    // Check if the user has privileges to pin the post
+    const currentUser = await User.findOne({
```
xoldyckk, line 100: You're querying database two times, just to get current user's data. You've already done it before with currentUserExists. Instead of just using it to check whether user exists, fetch the user object, check its existence there and then also make use of its data here for comparision.

src/resolvers/Mutation/createPost.ts
```diff
@@ -93,6 +94,38 @@ export const createPost: MutationResolvers["createPost"] = async (
...
+      _id: context.userId,
+    }).lean();
+    if (
+      !(
```
xoldyckk, line 104: This comparision is hard to read. Simplify it to this:-
```
const currentUserIsOrganizationAdmin = currentUser.adminFor.some(organzationId=> organizationId.toString() === args.data.organizationId))

if (currentUser.userType !== "SUPERADMIN" && !currentUserIsOrganizationAdmin){
  throw error.....
}
```

src/resolvers/Mutation/togglePinnedPost.ts
```diff
@@ -0,0 +1,105 @@
...
+    );
+  }
+
+  // Toggle the post's status for the organization
+  const organization = await Organization.findOne({
+    _id: post.organization,
+  }).lean();
+
+  const currentPostIsPinned = organization!.pinnedPosts.some(
+    (p) => p.toString() === args.id.toString()
```
xoldyckk, line 71: Change `p` to `postId`. Try to be explicit about variable names. Short names just create confusion in the long run.

src/resolvers/Mutation/togglePinnedPost.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+      {
+        $pull: {
+          pinnedPosts: args.id,
+        },
+      },
+      {
+        new: true,
+      }
+    ).lean();
```
xoldyckk, line 87: If you don't want the returned result for a database mutation you should not use `findOneAndUpdate`, just use `updateOne`.

Same goes for using `lean()`. It's only used to transform a mongoose document to plain old javascript object for returning as result. Not needed here.

xoldyckk:
@EshaanAgg I've thought about the user experience some more and how the post will be available on the client side. If the information whether a post is pinned or not needs to be available on each post instead of just on a particular page which only lists pinnedPosts, then using this approach of putting pinned posts ids in organization will make it slow to query for state of pinned boolean for each individual post. Since we'll have to create a seperate resolver for each post which just checks whether the postId exists in `pinnedPosts` of the corresponding organization, it'll be very slow.

If admins can pin posts from any place, whether admin dashboard or the mobile app itself, then the `pinned` information needs to be present within the post model itself to reduce database queries. This user experience far outweighs the cost of a stale `pinned` field on the post object. So, I guess we should just put the `pinned` boolean inside the post object.

EshaanAgg:
> @EshaanAgg I've thought about the user experience some more and how the post will be available on the client side. If the information whether a post is pinned or not needs to be available on each post instead of just on a particular page which only lists pinnedPosts, then using this approach of putting pinned posts ids in organization will make it slow to query for state of pinned boolean for each individual post. Since we'll have to create a seperate resolver for Post type named `pinned` which will just check whether the parent.id(postId) exists in `pinnedPosts` of the corresponding organization(parent.organization). And it'll have to be done for each post, it'll be very slow if someone is querying for a list of posts instead of just one post.
> 
> If admins can pin posts from any place, whether admin dashboard or the mobile app itself, then the `pinned` information needs to be present within the post object itself to reduce database queries. This user experience far outweighs the cost of a stale `pinned` field on the post object. So, I guess we should just put the `pinned` boolean inside the post object.

@xoldyckk I don't think so that we should be doing the same. The reason for the same is as on the mobile app, when the organizations' feed is loaded, it would make sense to make just one query to the organisation, and not make an individual query to filter out all the `pinned` posts from all the posts of the organization. Once the list of posts is loaded, the UI can pass that data down to each of the child post detail pages.

We shoud probably employ both these strategies, that is having pinnedPosts stored on the organisation model, as well as store the pinned status on each post individually, so that we can have performant rendering of the UI of pinned posts in the mobile app with a single request (no need to make a separate query for pinnedPosts, and orgDetails separately), and then use the pinned status on each post for the other concern of user experience as you explained.

xoldyckk:
> > @EshaanAgg I've thought about the user experience some more and how the post will be available on the client side. If the information whether a post is pinned or not needs to be available on each post instead of just on a particular page which only lists pinnedPosts, then using this approach of putting pinned posts ids in organization will make it slow to query for state of pinned boolean for each individual post. Since we'll have to create a seperate resolver for Post type named `pinned` which will just check whether the parent.id(postId) exists in `pinnedPosts` of the corresponding organization(parent.organization). And it'll have to be done for each post, it'll be very slow if someone is querying for a list of posts instead of just one post.
> > If admins can pin posts from any place, whether admin dashboard or the mobile app itself, then the `pinned` information needs to be present within the post object itself to reduce database queries. This user experience far outweighs the cost of a stale `pinned` field on the post object. So, I guess we should just put the `pinned` boolean inside the post object.
> 
> @xoldyckk I don't think so that we should be doing the same. The reason for the same is as on the mobile app, when the organizations' feed is loaded, it would make sense to make just one query to the organisation to fetch all of its details as well as the pinned posts, and not make an individual query to filter out all the `pinned` posts from all the posts of the organization. Once the list of posts is loaded, the UI can pass that data down to each of the child post detail pages.
> 
> We shoud probably employ both these strategies, that is having pinnedPosts stored on the organisation model, as well as store the pinned status on each post individually, so that we can have performant rendering of the UI of pinned posts in the mobile app with a single request (no need to make a separate query for pinnedPosts, and orgDetails separately), and then use the pinned status on each post for the other concern of user experience as you explained.
> 
> What do you think?

Yes, the second option is a bit better yes. But not a whole lot of performance gains.

Also, you're probably thinking that filtering organization posts based on `pinned` boolean will be slow, but it isn't the case. Since a boolean can only have two possible values sorting posts according to `pinned` field will be very fast. Databases are very smartly designed to handle these things.

Also, we should just index the organizationId field on post object. That way filtering the posts according to different organizations will be wayyy faster. It will not only be beneficial for `pinned` posts but also for querying posts for organizations by their foreign keys, whether the posts are `pinned` or not.

As a last resort, you can index the `pinned` field as well to achieve more speed but it'll not make much difference because it's just a boolean and can only have two distinct values at any time.

EshaanAgg:
Yes I agree that the gain in performance wouldn't be much, but atleast it would be less burden and more flexibility for clients. 
Thanks for the detailed discussion on the performance aspect, and I'll read about creating an index on the `organizationID` for the `Post` model so that they can be queried more optimally. 

I'll also update the PR to follow the joint approach, and resolve all the conversations that you have provided above.

src/typeDefs/types.ts
```diff
@@ -199,6 +199,7 @@ export const types = gql`
     apiUrl: URL!
     createdAt: DateTime
     tags: [String!]!
+    pinnedPosts: [Post]!
```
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/resolvers/Mutation/togglePinnedPost.ts
```diff
@@ -0,0 +1,105 @@
...
+  if (!post) {
+    throw new errors.NotFoundError(
+      requestContext.translate(POST_NOT_FOUND_MESSAGE),
+      POST_NOT_FOUND_CODE,
+      POST_NOT_FOUND_PARAM
+    );
+  }
+
+  // Check if the current user is authorized to perform the operation
+  const currentUser = await User.findOne({
```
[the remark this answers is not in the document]
EshaanAgg: Ahh makes sense. An absent minded mistake on my end. Will rectify the same.

src/resolvers/Mutation/createPost.ts
EshaanAgg: The current implementation and one you are suggesting are equivalent. With your approach, we would only get the true value if `true === true`, which is a redundant check, JS will treat `null`, `undefined` and `false` all to be `falsy` by deafult.

src/resolvers/Mutation/createPost.ts
EshaanAgg: Got it.

src/resolvers/Mutation/togglePinnedPost.ts
```diff
@@ -0,0 +1,105 @@
...
+    _id: context.userId,
+  }).lean();
+
+  if (
+    !(
+      currentUser!.userType === "SUPERADMIN" ||
```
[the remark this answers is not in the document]
EshaanAgg: Got it. Will do.

src/resolvers/Mutation/createPost.ts
EshaanAgg: Got it.

src/resolvers/Mutation/togglePinnedPost.ts
EshaanAgg: Got it.

src/resolvers/Mutation/togglePinnedPost.ts
EshaanAgg: Okay got it.

xoldyckk:
> Yes I agree that the gain in performance wouldn't be much, but atleast it would be less burden and more flexibility for clients. Thanks for the detailed discussion on the performance aspect, and I'll read about creating an index on the `organizationID` for the `Post` model so that they can be queried more optimally.
> 
> I'll also update the PR to follow the joint approach, and resolve all the conversations that you have provided above.

Flexibility for clients is expressed in graphql schema not database. How you efficiently resolve the data for the flexibility graphql is offering is the backend's concern.

EshaanAgg:
Getting your concerns @xoldyckk! Data duplication is as a general rule should be avoided, but here for our specific use case, I think we do have a bit of use for it. Probably not the most efficient implementation, but a good one nonetheless given our specific use case.

xoldyckk:
> Getting your concerns @xoldyckk! Data duplication is as a general rule should be avoided, but here for our specific use case, I think we do have a bit of use for it. Probably not the most efficient implementation, but a good one nonetheless given our specific use case.

OK

src/resolvers/Mutation/createPost.ts
```diff
@@ -93,6 +95,47 @@ export const createPost: MutationResolvers["createPost"] = async (
...
+      { _id: args.data.organizationId },
+      {
+        $push: {
+          pinnedPosts: createdPost._id,
+        },
+      }
+    );
+
+    // Change the pinned status of the post
+    await Post.updateOne(
```
xoldyckk, line 127: Arrange the checks in a way that `pinned` is set in the same step the post is created.

src/resolvers/Mutation/createPost.ts
```diff
@@ -93,6 +95,47 @@ export const createPost: MutationResolvers["createPost"] = async (
...
   });
 
+  // Check if the pinned property is set
+  if (args.data.pinned) {
+    // Check if the user has privileges to pin the post
+    const currentUserIsOrganizationAdmin = currentUser.adminFor.some(
+      (organizationId) => organizationId.toString() === args.data.organizationId
+    );
+
+    if (
```
xoldyckk, line 105: I don't think you should throw error when you've already created the post. Do it before creating the post. 

Or don't throw the error and instead, create the post but change the value of input `pinned` boolean to false, if the user who sent it is a normal user.

src/resolvers/Mutation/togglePinnedPost.ts
```diff
@@ -0,0 +1,124 @@
...
+  // Toggle the post's status for the organization
+  const organization = await Organization.findOne({
+    _id: post.organization,
+  }).lean();
+
+  const currentPostIsPinned = organization!.pinnedPosts.some(
+    (postID) => postID.toString() === args.id.toString()
+  );
+
+  let updatedPost: Interface_Post | null = null;
```
xoldyckk, line 72: Since findOneAndUpdate returns the updated post, return it where you're updating it. Creating updatePost variable is useless here.

`return await Post.findOneAndUpdate....`

src/resolvers/Mutation/createPost.ts
EshaanAgg: In the latest push, I am checking to see if the user is authorized to create the same or not. If not, Ill throw the error, otherwise I continue to make the post after that.

src/resolvers/Mutation/createPost.ts
```diff
@@ -85,14 +87,44 @@ export const createPost: MutationResolvers["createPost"] = async (
...
+        requestContext.translate(USER_NOT_AUTHORIZED_TO_PIN.message),
+        USER_NOT_AUTHORIZED_TO_PIN.code,
+        USER_NOT_AUTHORIZED_TO_PIN.param
+      );
+    }
+  }
+
+  // Creates new post
   const createdPost = await Post.create({
     ...args.data,
```
xoldyckk, line 110: What if the `pinned` variable is null, undefined? We should not allow those values in the database. Or even it they're not allowed, this should not result in resolver failing to create a post. Check how mongodb/mongoose deals with it and whether we need to transform the input before storing it.

tests/resolvers/Mutation/togglePinnedPost.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+    const { togglePinnedPost: togglePinnedPostResolver } = await import(
+      "../../../src/resolvers/Mutation/togglePinnedPost"
+    );
+
+    await togglePinnedPostResolver?.({}, args, context);
+    const organization = await Organization.findOne({
+      _id: testPost!.organization,
+    }).lean();
+
+    const currentPostIsPinned = organization!.pinnedPosts.some(
```
xoldyckk, line 180: You also have to check that on the actual post, the boolean is toggled.

src/resolvers/Mutation/createPost.ts
EshaanAgg: Used a check in the latest commit to only restrict it to true/false

tests/resolvers/Mutation/togglePinnedPost.spec.ts
EshaanAgg: Updated the tests

EshaanAgg:
Got it @xoldyckk! Superadmin should also have access to do so right?

xoldyckk:
> Got it @xoldyckk! Superadmin should also have access to do so right?

IDK tbh.

palisadoes:
> Got it @xoldyckk! Superadmin should also have access to do so right?

Yes. Super Admins should be able to do whatever an Admin can do for their organization

EshaanAgg:
Yes @palisadoes, will do so as soon as all the changes in this PR are approved. Would make a new PR in `talawa-docs` and add reference to the original issue in the same.

src/resolvers/Mutation/togglePinnedPost.ts
```diff
@@ -0,0 +1,121 @@
+import {
```
xoldyckk, line 1: rename the function to `togglePostPin`. Everything else looks good.

EshaanAgg: Done.

xoldyckk (approved):

palisadoes (approved):

Issue #1101, opened 2023-03-02
Unable to Send Create Event Mutation


anshgoyalevil:
**Describe the bug**
The Create Event Mutation is not Working in Talawa-Admin, due to some bug in Talawa-API. 

**To Reproduce**

1. Go to Talawa-Admin
2. Open Any Org
3. Go to Events Page and try creating any Event
4. You will get 400 response code error

**Expected behavior**
The event should have been created

**Actual behavior**
An error is occurring right now. 

**Screenshots**
![image](https://user-images.githubusercontent.com/94157520/222318891-8b090215-c99e-4e09-b8b2-f28e83f6d08b.png)

The response received here is 
```
{"errors":[{"message":"Variable \"$startDate\" got invalid value \"Wed Mar 01 2023\"; Date cannot represent an invalid date-string Wed Mar 01 2023.","status":422,"data":[]},{"message":"Variable \"$endDate\" got invalid value \"Thu Mar 02 2023\"; Date cannot represent an invalid date-string Thu Mar 02 2023.","status":422,"data":[]},{"message":"Variable \"$startTime\" got invalid value \"\"; Time cannot represent an invalid time-string .","status":422,"data":[]},{"message":"Variable \"$endTime\" got invalid value \"\"; Time cannot represent an invalid time-string .","status":422,"data":[]}]}
```

**Additional details**
Due to this, the events functionality is broken for now.

anshgoyalevil:
Please assign this issue to me.

EshaanAgg:
@anshgoyalevil This is due to introduction of GraphQL-Scalars into the API. I had also discovered the same while coordinating with @literaleval for Talawa migration. 

The error is arising due to the fact that the Event model uses String type to store these parameters in the database.

EshaanAgg:
I had already implemented a bug fix locally at https://github.com/eshaanAgg/talawa-api/tree/scalar_fix so that I could co-ordinate with @literaleval. 

Can you confirm once if the using the same fixes the error or not?

anshgoyalevil:
@EshaanAgg My solution would look something like that only. Should I close this issue for now? Or you need to be assigned to this issue.

EshaanAgg:
If using the same is fixing the issue and if it's alright with you, I would like to raise the PR for this issue as I have already code ready (which has been tested with talawa-mobile as well)

anshgoyalevil:
[...]

PR #1103, opened 2023-03-02
Change the data type of storage of attributes of Event Model [Fixes part of #1101]

EshaanAgg:
Bugfix

**Issue Number:**

Fixes part of #1101 
Must be merged with [this PR](https://github.com/PalisadoesFoundation/talawa-admin/pull/523) to restore the functionality

NA

**Snapshots/Videos:**
NA

NA

**Summary**
Updated the `endDate`, `startDate`, `endTime` and `startTime` variables to be stored as Date type in the database (and not String) so that they can be serialized correctly by `GraphQL-Scalars`
This code has been tested with talawa-mobile's create event functionality too. 

No

**Other information**
NA

Yes

### DIFF
--- a/src/models/Event.ts
+++ b/src/models/Event.ts
@@ -91,25 +91,25 @@ const eventSchema = new Schema({
     required: true,
   },
   startDate: {
-    type: String,
+    type: Date,
     required: true,
   },
   endDate: {
-    type: String,
+    type: Date,
     required: function (): () => boolean {
       // @ts-ignore
       return !this.allDay;
     },
   },
   startTime: {
-    type: String,
+    type: Date,
     required: function (): () => boolean {
       // @ts-ignore
       return !this.allDay;
     },
   },
   endTime: {
-    type: String,
+    type: Date,
     required: function (): () => boolean {
       // @ts-ignore
       return !this.allDay;


palisadoes (approved):

xoldyckk (approved):

xoldyckk:
@palisadoes @EshaanAgg Dependence on `allDay` boolean is weird if you ask me. Also, the api and database don't handle the edge cases well. Edge cases here bascially are all the combinations of fields like startTime, endTime, startDate, endDate, recurrance and what their contextual meaning is.

xoldyckk:
There could be some inconsistencies with accepting `Date` and `Time` seperately as without the complete `DateTime` the date conversion will be incorrect. At least that's what I think, I don't know much about all the intricacies with date times as it's a broad field.

EshaanAgg:
@xoldyckk I feel that too. Instead of managing these four fields separately, we should rather be using just two variables `start` and `end`, each of `DateTime` type. Though implementing the same would be a breaking change, and would require more review before proceeding with the same.

xoldyckk:
> @xoldyckk I feel that too. Instead of managing these four fields separately, we should rather be using just two variables `start` and `end`, each of `DateTime` type. Though implementing the same would be a breaking change, and would require more review before proceeding with the same.

Also, you've annotated mongodb field as Date, but what you're giving it is just a `Date` and `Time`. Does it automatically convert it to its UTC equivalent?

xoldyckk:
> MongoDB has no support for Date and Time types separately, but rather only for storing DateTime under the type name `Date`. It the the `GraphQL-Scalars` package that does this conversion (and reverts the same upon fetching) to the `UTC` equivalent before storage in the database,

Have you tested this manually? Whether the input fields like `startDate`, `startTime` are converted to their UTC equivalent then given to resolvers as input?

palisadoes:
"All Day" is just a UX convenience. Storing the event timing data as a start/stop only should be OK. We can just let the other repos know that the "All Day" feature is going to have to be refactored in their code.

How do you propose to handle "recurring" events as there is going to be a lot of meta data? (Google calendar has lots of recurring event options both custom and standard"

palisadoes:
Fixing event timing should be handled as a separate issue, and we should merge this PR. I don't want to get into another discussion over architecture and delay the necessary fix this PR will provide.

@xoldyckk can you create the fix events issue so that this can be merged?

xoldyckk:
> Fixing event timing should be handled as a separate issue, and we should merge this PR. I don't want to get into another discussion over architecture and delay the necessary fix this PR will provide.
> 
> @xoldyckk can you create the fix events issue so that this can be merged?

Yes, this can be merged. Because the logic of storing it as DateTime in mongodb is correct I guess.

EshaanAgg:
> > MongoDB has no support for Date and Time types separately, but rather only for storing DateTime under the type name `Date`. It the the `GraphQL-Scalars` package that does this conversion (and reverts the same upon fetching) to the `UTC` equivalent before storage in the database,
> 
> Have you tested this manually? Whether the input fields like `startDate`, `startTime` are converted to their UTC equivalent then given to resolvers as input?
> 
> If you haven't you can console log it inside the resolver while creating the event and see.

Yes I have tested the same. Rather than testing in the backend, I have tested the same of the frontend by logging the same.
Date Type Scalar returns the date in the form of `YYYY-MM-DD` while Time Type Scalar returns the same in the form of `HH:MM:SS.000Z`, and thus the UI has been changed to handle it.

PR #1110, opened 2023-03-03
MongoDB (Mongoose) validation for valid String 

### LINKED ISSUE #1097: Security : Database Input Validation for String Validity (DDL)
**Describe the bug**
Right now, certain checks are implemented for user input coming from the client app in the resolver. 

However, those constraints do not exist in the data model which could lead to unexpected Cross-site-Scripting and Injection attacks as the database mutation can be added without passing through the API layer.

A more detailed explanation can be found in the discussion between @anshgoyalevil and I. here https://thepalisadoes-dyb6419.slack.com/archives/CSWH4PN0M/p1677490069884249?thread_ts=1677432883.833189&cid=CSWH4PN0M
 
**To Reproduce**

1. Start the mongo server /  open the atlas console
2. Use your instance of `talawa-api` db  
3. Insert object in a user database with a password of length one or insert a string with length>300 characters.
4. The data is inserted without any errors.

**Expected behavior**
The object should not be inserted and should throw an error
**Actual behavior**
The invalid data is inserted without any errors.
**Screenshots**

**Additional details**
The issue can be resolved very easily with mongoose validators which under the hood add applied constraints within the collection like a DDL for RDBMS. SO that logically invalid or malicious data does not get inserted into the DB.

kb-0311:
This PR added mongoose validation for the mongoDB validators to make sure invalid mutations are not injected into the database.

**Issue Number:**

Fixes #1097

Yes.

**Snapshots/Videos:**

**Summary**

**Other information**


kb-0311:
This PR was getting too big so I will be validating the dates in later pr so that it is easier to review.

palisadoes:

kb-0311:
@palisadoes fixed the git conflicts.

noman2002 (dismissed):
Looks good to me.

xoldyckk:
@kb-0311 What is the need for this?

xoldyckk:
@kb-0311 @anshgoyalevil 

I don't understand the talk about nosql injections. Most ORMs already protect you from such things if you're interacting with the database using their provided methods. Injections mostly only happen when you're using raw sql/nosql, something like using mongodb nodejs driver.

There's also a package named `mongo-sanitize` being used for exactly this purpose. It's technically just double protection.

The talk about accessing the database directly and manipulating values, that never happens. Why would a database administrator makes changes to values, that have their business logic defined in api layer? It's not like you can define every single business logic inside mongodb itself. Any DBA who accesses database would mostly do so for querying, managing the database. They'd never mutate records directly. That would be stupid. If they absolutely must do it, they'd need to have to know the business logic in api layer, or they'd need assistance from someone who does. Also, putting things like min, max value in database is stupid, database doesn't dictate it, api does. Databases have a simple job, store data in an organized way. Of course there are some constraints which should be put in the database, something like uniqueness of a field, or type of a field(date for example), enums etc. Databases have special meaning for these things and they provide certain additional functionalities for them.

Using custom regexes is just impending disaster. You should never do it, unless you're exceptionally skilled in them. Maintaining them is the biggest pain. Also, they're ugly.

The most important thing, those regexes only allow certain characters in the database. Talawa is supposed to be accessible by users of multiple languages, but you're limiting them to use only english. Javascript strings support utf8 characters, there can be many languages, emojis, special characters etc., but you're technically stripping down all those to only support ASCII.

anshgoyalevil:
@xoldyckk 

Yes, I was having the same points. Here's the Slack Discussion for that: https://thepalisadoes-dyb6419.slack.com/archives/CSWH4PN0M/p1677432883833189

I just over secured the API resolvers using Regex, and those will restrict the users with language other than English. 
I will create an issue, and remove the pattern validation for strings in mutation resolvers all over the Talawa-API.

kb-0311:
@xoldyckk 

>There's also a package named mongo-sanitize being used for exactly this purpose. It's technically just double protection.

I would have to disagree with you a bit on this one, chief. I could not find if mongoose protects from noSQL attacks by default  

`mongo-sanitize` package does not prevent me from inserting invalid mutation / malicious code (according to API logic) from a mongo shell let's say for a defined model, it only sanitizes the input we are passing as variables in the API resolver layer. for eg , I could still `db.users.create()` my way through that.

>The talk about accessing the database directly and manipulating values, that never happens. Why would a database administrator makes changes to values, that have their business logic defined in api layer? It's not like you can define every single business logic inside mongodb itself. Any DBA who accesses database would mostly do so for querying, managing the database. They'd never mutate records directly. That would be stupid. If they absolutely must do it, they'd need to have to know the business logic in api layer, or they'd need assistance from someone who does.

That is how it _should_ be, I don't think it is how it always one hundred percent happens without any error whatsoever, guaranteed. My opinion here is if someone could do it, we should protect the possibility of something going wrong as much as possible. I guess your opinion may differ here and it would just be a pointless debate.

I don't think you are that opposed to the idea of double protection just what kinda double protection we are adding at the mongoose layer. 

>Also, putting things like min, max value in database is stupid, database doesn't dictate it, api does. Databases have a simple job, store data in an organized way. Of course there are some constraints which should be put in the database, something like uniqueness of a field, or type of a field(date for example), enums etc. Databases have special meaning for these things and they provide certain additional functionalities for them.

I don't know why you would think uniqueness of a field has special meaning to the ODM we are using in mongoDB since mongoose does not even support the unique check https://mongoosejs.com/docs/validation.html#the-unique-option-is-not-a-validator. For that we will need a custom validator exactly like in this PR.

I also don't understand why you think a min-max validation is dumb in the database (a lot of apps I have seen use them), even if we are talking about the development process a developer working locally, without changing the code, a dev can `db.user.create()` his way into getting inserted as `email : '"1@2.com" , password "ok"` and can successfully login from the admin app or API(since we do not validate login inputs) , all of that without changing a single line of code in their local instance, you don't think that sus?

Ofc, I agree not every single business logic needs to be validated tho


There are also other benefits, now I am a total beginner in API development, no way I am gonna read through 10,000 lines of resolver code to understand business validations. If I can just glance over the models to get a rough understanding of them.

>I thought I had said this before but I'll say it again, using custom regexes is just impending disaster. You should never do it, unless you're exceptionally skilled in them. Maintaining them is the biggest pain. Also, they're ugly.

>The most important thing, those regexes only allow certain characters in the database. Talawa is supposed to be accessible by users of multiple languages, but you're limiting them to use only english. Javascript strings support utf8 characters, there can be many languages, emojis, special characters etc., but you're technically stripping down all those to only support ASCII.

Wow, I never actually thought of that. I agree about this. I will remove the regex checks for this PR tomorrow.

palisadoes:
@kb-0311 please also verify whether we are using regress in any other parts of the code base for Talawa API and Admin, and create issues when found. I never thought about the language problem either.

Kevoniat (dismissed):

anshgoyalevil:
@xoldyckk 

Should we use any other validator for string pattern or ignore that completely?

palisadoes:
@xoldyckk Can we merge this now? It's been approved twice.

kb-0311:
@palisadoes pls don't merge it rn wait for @xoldyckk 's feedback plus I also need to remove the regex check here.

There are regex checks put in places in API resolver by @anshgoyalevil and he said he is gonna remove that too.

palisadoes:
Understood, but we are having scope creep.

1. Do we need to wait for the removal of regexes for other resolvers?
1. Shouldn't that be another PR especially as they don't appear to be breaking changes yet? It seems to be out of scope for this PR.
1. @anshgoyalevil what is the issue # for the regex removal for other resolvers? We'll need one so that we can track progress.

anshgoyalevil:
@kb-0311 @palisadoes 

I am creating an issue now for removal of regexes in resolvers.

@kb-0311 You proceed with removing the regexes from Database Layer. I will handle the API Resolvers Layer.

kb-0311:
@palisadoes yes, the removal of regex check is another PR as mentioned by @anshgoyalevil above.

xoldyckk:
> @xoldyckk
> 
> > There's also a package named mongo-sanitize being used for exactly this purpose. It's technically just double protection.
> 
> I would have to disagree with you a bit on this one, chief. I could not find if mongoose protects from noSQL attacks by default
> 
> `mongo-sanitize` package does not prevent me from inserting invalid mutation / malicious code (according to API logic) from a mongo shell let's say for a defined model, it only sanitizes the input we are passing as variables for db operations in the API resolver layer. for eg , I could still `db.users.create()` my way through that.
> 
> > The talk about accessing the database directly and manipulating values, that never happens. Why would a database administrator makes changes to values, that have their business logic defined in api layer? It's not like you can define every single business logic inside mongodb itself. Any DBA who accesses database would mostly do so for querying, managing the database. They'd never mutate records directly. That would be stupid. If they absolutely must do it, they'd need to have to know the business logic in api layer, or they'd need assistance from someone who does.
> 
> That is how it _should_ be, I don't think it is how it always one hundred percent happens without any error whatsoever, guaranteed. My opinion here is if someone could do it, we should protect the possibility of something going wrong as much as possible. I guess your opinion may differ here and it would just be a pointless debate.
> 
> I don't think you are that opposed to the idea of double protection just what kinda double protection we are adding at the mongoose layer.
> 
> > Also, putting things like min, max value in database is stupid, database doesn't dictate it, api does. Databases have a simple job, store data in an organized way. Of course there are some constraints which should be put in the database, something like uniqueness of a field, or type of a field(date for example), enums etc. Databases have special meaning for these things and they provide certain additional functionalities for them.
> 
> I don't understand why you would say uniqueness has a special meaning for our ODM for our mongoDB when mongoose does not even support it by default https://mongoosejs.com/docs/validation.html#the-unique-option-is-not-a-validator. For that we would need a custom validator exactly like the one in this PR. https://mongoosejs.com/docs/faq.html#unique-doesnt-work I also don't understand why you think a min-max validation is dumb in the database (a lot of apps I have seen use them), even if we are talking about the development process a developer working locally, without changing the code, a dev can `db.user.create()` his way into getting inserted as `email : '"1@2.com" , password "ok"` and can successfully login from the admin app or API(since we do not validate login inputs , and rightfully so) , all of that without changing a single line of code in their local instance, you don't think that sus?
> 
> Ofc, I agree not every single business logic needs to be validated tho
> 
> There are also other benefits, now I am a total beginner in API development, no way I am gonna read through 10,000 lines of resolver code to understand business validations. If I can just glance over the models to get a rough understanding of them.
> 
> > I thought I had said this before but I'll say it again, using custom regexes is just impending disaster. You should never do it, unless you're exceptionally skilled in them. Maintaining them is the biggest pain. Also, they're ugly.
> 
> > The most important thing, those regexes only allow certain characters in the database. Talawa is supposed to be accessible by users of multiple languages, but you're limiting them to use only english. Javascript strings support utf8 characters, there can be many languages, emojis, special characters etc., but you're technically stripping down all those to only support ASCII.
> 
> Wow, I never actually thought of that. I agree about this. I will remove the regex checks for this PR tomorrow.
> 
> Btw do you think that adding an extra layer of protection here is unnecessary or the reasons I am giving for implementing them are not that impactful?

If mongoose doesn't prevent against nosql injections, that just means it should be ditched, cuz it's the job of the orm to provide such functionality. But I've seen no talk about nosql injections with mongoose so idk where you got that information from. If it was such a serious issue mongoose would itself document it in their docs. Also, you should look more into how nosql injections work, and whether they're feasible with how talawa-api currently accepts user input and how those inputs are inserted into the database. Injections mostly only occur when inputs become a part of db queries instead of just placeholders for values. When they're parsed by db engines, the engine executes those inputs as a part of query instead of input. That's usually how it goes.

Mongo sanitize does prevent against nosql injections. From their docs:-
```
// By default, $ and . characters are removed completely from user-supplied input in the following places:
// - req.body
// - req.params
// - req.headers
// - req.query
```
If you're accepting user input by just letting them pass in stringified json which could be valid mongodb query syntax, then yes that's could be a nosql injection and that too will only happen if you directly make it a part of query(something like javascript's template literals `${input.username}`. If mongo sanitize detects this kind of syntax in input, it'll purge them. And graphql args and inputs are the same thing, they're passed in inside req.body.

I'm not against putting constraints in database, but there are times when a decision affects the application in long run. For example, you make maximum length of a description field to be 400. Later comes a time when you need to reduce it down to 300. What then? What happens to pre-existing data? Whereas, something like a credit card number max/min length is a good thing to put as database constraint as it will always be the same length no matter what.

There's also the thing about duplicating checks,constraints and validations at different layers, yes it's a good thing, having validation at as many places as it can be makes the application more robust and less error prone. But server should always be the first layer for validation. You don't want to waste a round trip to database just to validate stuff, it should just be validated in the server and appropriate errors should be returned to clients.

The thing about mongoose not upholding unique constraint violation, it's stated that it's not a problem in production, so I guess we just need to fix those in tests. It's a weird behaviour and I guess I have another reason to hate mongoose now lol.

You're saying a dev can just create a object using without validation, yeah let them, they're doing it in development, they can do whatever the hell they want with their database. Of course the same won't be true when writing tests. The tests create data according to business logic they're testing, not by the whims of the developer writing those tests. And yes, those tests will need to be thoroughly reviewed by mentors.

The way you're handling those errors is very confusing and naive. Though I myself am not expert at code organization and architecture, I think for the current use case I can provide some thoughts for clean code. First thing to note is the layers in current code. 

A good standard server architecture would look like this:-
```
api protocols(rest graphql grpc etc.) layer
business logic layer
database(persistence) layer
```
Mongoose can be thought of as database layer here.

Let's focus on graphql here. The only job for graphql resolvers in such an architecture would be to call functions defined in business layer, pass input/receive output. Now the output they receive would either be a success(a result) or a failure(errors). If errors were returned we throw appropriate graphql errors within resolvers by transforming those according to graphql conventions. If results were returned we pass the results further in the resolver(either to return to client or make use of in some other business logic functions).

Talawa-api is currently a monolith, it has no such layers. It's strictly graphql, so there's no need for seperate business logic layer. All business logic can exist inside resolvers themselves and errors/result can be returned. Previously we had business logic layer return those errors. Now we'll have to validate/transform user input within resolver itself. Now that's not to say that they can't be extracted into their own functions. Those extracted functions will be easy to unit test and make the code very clean. This isn't only limited to user input, there are many other instances of checks within resolvers which are duplicated, we can also extract them out as seperate functions.

There's also some use cases where it would be better to rely on database errors instead of manual validation, one such case is unique constraint violation. We'd wrap our resolvers in try catch blocks whereever it's necessary to catch database errors relevant to resolvers. We can check those errors for the type of error we're looking for, if it's found we return to the client a nicely formatted graphql error. If not, we let the error propagate to higher layers, where it would be handled using a function like apollo server's formatError, it'd throw something like `internal sever error` cuz client doesn't need to know about those errors.

kb-0311:
@xoldyckk ohk I got where you are coming from. 

1. The regex check was the thing for injection attacks and it will be getting removed. 
2. Ahh alright you are saying you are not opposed to an added protection of database validation, if in the future we have multiple services other than graphql server accessing the db then we could return these mongoDB instance errors to the client. 
3. But the main validation layer would still have to be in the resolvers and preferably be developed as their own functions.

That seems okay.

I would still say keep the length check for this PR, as an added validator.
>For example, you make maximum length of a description field to be 400. Later comes a time when you need to reduce it down to 300. What then? What happens to pre-existing data? Whereas, something like a credit card number max/min length is a good thing to put as database constraint as it will always be the same length no matter what.

Thankfully adding validation to an existing collection does not enforce validation on existing documents. So they will remain unaffected.

https://www.mongodb.com/docs/manual/core/schema-validation/#when-mongodb-checks-validation

The main idea of the PR was to enforce these checks in the database layer so that there would exist at least some consistency of the DB operations with the API validation logic. 

>The way you're handling those errors is very confusing and naive. Though I myself am not expert at code organization and architecture, I think for the current use case I can provide some thoughts for clean code.

If at any point in a service, we would need to rely on `MongooseInstanceErrors` then what should be our approach there? What changes can be made to how I am handling them here?

kb-0311:
@palisadoes could you re-request a review from kevonia and noman?

palisadoes:
Is it ready to be merged?

xoldyckk:
> @xoldyckk ohk I got where you are coming from.
> 
> 1. The regex check was the thing for injection attacks and it will be getting removed.
> 2. Ahh alright you are saying you are not opposed to an added protection of database validation, if in the future we have multiple services other than graphql server accessing the db then we could return these mongoDB instance errors to the client.
> 3. But the main validation layer would still have to be in the resolvers and preferably be developed as their own functions.
> 
> That seems okay.
> 
> I would still say keep the length check for this PR, as an added validator.
> 
> > For example, you make maximum length of a description field to be 400. Later comes a time when you need to reduce it down to 300. What then? What happens to pre-existing data? Whereas, something like a credit card number max/min length is a good thing to put as database constraint as it will always be the same length no matter what.
> 
> Thankfully adding validation to an existing collection does not enforce validation on existing documents. So they will remain unaffected.
> 
> https://www.mongodb.com/docs/manual/core/schema-validation/#when-mongodb-checks-validation
> 
> The main idea of the PR was to enforce these checks in the database layer so that there would exist at least some consistency of the DB operations with the API validation logic.
> 
> > The way you're handling those errors is very confusing and naive. Though I myself am not expert at code organization and architecture, I think for the current use case I can provide some thoughts for clean code.
> 
> If at any point in a service, we would need to rely on `MongooseInstanceErrors` then what should be our approach there? What changes can be made to how I am handling them here?

Read again. I'm saying putting some constraints in database are non-sensical if they're bound to easily change, instead you should use a generic type in database and enforce constraints in api. Yes, mongodb will allow you to freely add/remove constraints in database and that's why it's one of the most hated databases because it allows us to do anything with our data. Another thing is the more constraints and checks you put in database the slower they'll become.

No, I'm not saying that these mongo errors would be used for api protocols. I was giving an example of what a business logic layer is. User input with graphql could be different than the user input with rest api . It's the job of api protocol layer to transform those different inputs  to conform to the shape business logic layer wants them in.  Api protocol layers will only feed the data to business layer, all the validation/constraint check and db operations would be handled by business layer. Then it'll return either errors or result and your api protocol layer will handle them accordingly by transforming into the kinds of responses they want to send to client(for rest api you send 404 not found, for graphql you send `{data:null, errors:[....not found error]}` etc.).

Technically you can do all the constraint/validation checks within the business layer. But it is not efficient, if you need data from database. Like in this case the business layer can't know beforehand whether a user exists with username or email. It has to query the database to get that information. So, the approach below makes multiple database calls to check for unique constraint check. This is a bad approach as it leads to multiple round trips to database.
```
try(
 const existingUserWithUsername = await db.user.findOne({
     username: input.username
   })

   if(existingUserWithUsername){
      return usernameTakenError
   }

  const existingUserWithEmail = await db.user.findOne({
     username: input.email
   })
   
   if(existingUserWithEmail){
     return emailTakenError
   }
   
  const createdUser = await db.user.create({
  username:input.username,
  email:input.email
  })
  
  return createdUser
)catch(e){
 ...send unexpected error to client
}
```

A much better approach is to rely on database for these checks. There's always gonna be only one database call, and database can throw error accordingly or create the data if the constraints are fullfilled. All this with one database call. But this is done mildly, we don't want to rely on database for everything and write custom db error parsers to extract out errors for different constraints and validations. 
```
try{
  const createdUser = await db.user.create({
  username:input.username,
  email:input.email
  })
}catch(e){
  if(e instance of MongooseError){
    e.uniqueConstraintViolations.map(v=>{
      if(v === "UsernameTaken"){
        return usernameTakenError
      }else if(v === "EmailTaken){
         return EmailTakenError  
      }
    })
  }
  
 ...send unexpected error to client
}
```

So, we'd be doing the same thing but instead of business logic layer we'll do all this within the graphql resolver as currently there's no need for such complex structure. 

Also, the thing about understanding business logic by looking at mongoose models, that's another weird take. You understand business logic by reading app documentation and the test code, not by looking at database models.

Well anyway here's my final take. If you put all business logic only in mongoose schema, you're tying your app to only use mongoose and mongodb for the forseeable future. If this is ok then do so. If you're putting constraints/validations in both resolvers and mongoose schema there's two places the business logic needs to be kept in sync with each other and maintained continually. If this route is chosen, there should be an upfront decision about the constraints your putting in database(not that it matters with mongodb lol) so they aren't required to be changed for the forseeable future.

kb-0311:
@xoldyckk ok got it. The syncing part could lead to potential slip-ups during some other PRs too and could introduce inconsistencies.

@palisadoes closing this issue for now

PR #1120, opened 2023-03-05
First user default super admin

### LINKED ISSUE #1119: Promote user to super admin based on email in configuration
**Describe the bug**
More information on this issue - [issue](https://github.com/PalisadoesFoundation/talawa-admin/issues/499#issue-1597263725)

BhuvaneshPatil:

It promotes the user to SUPER ADMIN when it's registered with the configured email.

Feature

**Issue Number:**

resolves #1119 , resolves PalisadoesFoundation/talawa-admin#499

Yes. Added two test. First is to test if the user with the email that is configured to be last resort super admin email is promoted. And second is to test if other email are not promoted.

**Summary**
Before these changes the user had to manually change the property in the database for promoting the user to be Super Admin. Now we don't need that,

No

Yes


anshgoyalevil (approved):
Looks good!

Cioppolo14:
@BhuvaneshPatil Please refer to the PR_GUIDELINES.md document in the root folder of the repo. It tells you how to automatically close your issue when the PR is merged.  Also, please resolve the failed tests.

BhuvaneshPatil:
> @BhuvaneshPatil Please refer to the PR_GUIDELINES.md document in the root folder of the repo. It tells you how to automatically close your issue when the PR is merged. Also, please resolve the failed tests.

My tests are passing locally, the test uses env variable, do we have to perform any different action for those test cases?
image

.github/workflows/pull-request.yml
```diff
@@ -74,6 +74,7 @@ jobs:
       MONGO_DB_URL: mongodb://localhost:27017/talawa-test-db
       ACCESS_TOKEN_SECRET: eeb2137b1961a3da9e3736ea4a9ed1e77cb33a4fad8a153d364c2f04ec50eab7
       REFRESH_TOKEN_SECRET: 96d41e144e63475f7f753735b50305cd2857cc41c2f2570babdb948c10b77bc5
+      LAST_RESORT_SUPERADMIN_EMAIL: resortadmin@emailcom
```
palisadoes, line 77: Why did you add this here? It's it requires for testing?

palisadoes (changes requested):
1. Please fix the things I requested.
1. Please answer my questions.

INSTALLATION.md
```diff
@@ -112,6 +112,7 @@ This `.env` file must be populated with the following environment variables for
 | RECAPTCHA_SECRET_KEY | Used for authentication using reCAPTCHA                |
 | MAIL_USERNAME        | Used for mailing service                               |
 | MAIL_PASSWORD        | Used for mailing service                               |
+| LAST_RESORT_SUPERADMIN_EMAIL | Used for promoting default super admin         |
```
palisadoes, line 115: Provide more information about this parameter. The person installing the app needs to know why it's important not just during installation but also afterwards

palisadoes (changes requested):
Please see my comments and resolve them

palisadoes:
We are almost there. We need these few changes and it'll greatly improve the experience of the end user. Please see my comments

INSTALLATION.md
```diff
@@ -199,6 +201,12 @@ For mongosh info see: https://docs.mongodb.com/mongodb-shell/
 
 ```
 
+# Setting up .env LAST_RESORT_SUPERADMIN_EMAIL parameter
```
palisadoes, line 204: This needs to be ## not #

palisadoes (changes requested):
Please make the requested changes. This is almost complete

INSTALLATION.md
BhuvaneshPatil: sorry, will do the change now

palisadoes (approved):

BhuvaneshPatil:
The test is failing because of env variables in  production

palisadoes:
How can this be fixed? It will fail all the time for all other PRs if not corrected

anshgoyalevil:
@BhuvaneshPatil 

You can mock the Constant variables using Vitest. We don't need to change .env variables for PR workflow.

BhuvaneshPatil:
Thank you @anshgoyalevil , I made the changes

palisadoes:
Please modify the first comment so that the issue autocloses on merge. The PR_GUIDELINES.md file shows you how.

src/resolvers/Mutation/signUp.ts
```diff
@@ -112,6 +114,10 @@ export const signUp: MutationResolvers["signUp"] = async (_parent, args) => {
     password: hashedPassword,
   });
 
+  if (createdUser.email === LAST_RESORT_SUPERADMIN_EMAIL) {
```
xoldyckk, line 117: Create a variable createdUser, store in it result of creating a user. Create the user conditionally(checking whether the email is LAST_RESORT_SUPERADMIN_EMAIL), and assign the value to createdUser.

src/utilities/superAdminCreation.ts
```diff
@@ -0,0 +1,8 @@
+import { User } from "../models";
+
+export const updateUserToSuperAdmin = async (email: string) => {
```
xoldyckk, line 3: A seperate function just for this is wasteful. Remove this, and do this while creating the user itself.

BhuvaneshPatil: My idea for creating a separate function - 

Sign up function will mostly be used when user is created, by adding conditions to it we are making it bulky and adding extra responsibility. On the other hand if we put that logic in separate function it the responsibility is shared among them and in future if we add any other fields particularly for super admin we won't need to change the sign up function. 
Let me know your thoughts on the same.

xoldyckk: @BhuvaneshPatil But there is no logic being duplicated here. What's the point of having one utility function that's used at exactly one place? Besides with this approach you're using two database calls to do the same thing, why waste a database call?

If the same logic is duplicated 3 or more times then it makes sense to seperate out into a seperate function.

src/resolvers/Mutation/signUp.ts
```diff
@@ -104,12 +105,16 @@ export const signUp: MutationResolvers["signUp"] = async (_parent, args) => {
     uploadImageFileName = await uploadEncodedImage(args.file, null);
   }
 
+  const isLastResortAdmin = args.data.email === LAST_RESORT_SUPERADMIN_EMAIL;
```
xoldyckk, line 108: Change this to `isLastResortSuperAdmin`.

xoldyckk (dismissed):

palisadoes:
@BhuvaneshPatil

What is the default behavior if the `LAST_RESORT_SUPERADMIN_EMAIL` parameter is not configured? For example,
    1.  It remains as `LAST_RESORT_SUPERADMIN_EMAIL=` in the configuration file? People could forget to update it.
    1. The parameter is completely missing from the configuration. They were not paying attention to the INSTALATION.md file

In these cases:

1. Will the app crash?
1. What default behavior should we expect?

This is an important feature that automatically provides increased authorization. We don't want it incapacitating the application.

BhuvaneshPatil:
Hello @palisadoes . Sorry for getting back to you late.
Answers-
First Section - 

1. It will take an empty string as value `""` so it will compare the empty string with the email, which will evaluate to false.
2. In second case if it's not mentioned, it will be `undefined` , so again it will not cause any error. It will compare the email with `undefined ` that will be false.

Second Section - 

1. App won't crash in any of the situation.
2. In my opinion, the we should let the user do configuration, because it's one time thing. Some user may want to remove the `LAST_RESORT_ADMIN_EMAIL` once it's created. And we will highlight the point of setting this env variable in the INSTALLATION.md file.

If we decide to force user to set the env variables and make it a compulsory thing, 
We can do the following - 
Whenever server starts we can check if the value for email is set in env file and is non empty, if it's not we will prompt the user to set the env variable else won't start the server.

But this approach might fail, when the many admins are using the talawa-api using shared database locally.

Let me know you thought about the same.

cc - @xoldyckk

palisadoes:
1. Please update the documentation page that it's advisable to remove the email address after the initial installation and restore it only in emergencies.
1. We can then have a new issue to raise a warning if the setting is found in the file.

Let me know when the document is updated then we can merge. This is needed quickly.

BhuvaneshPatil:
yes working on it

palisadoes (approved):

PR #1130, opened 2023-03-06
Feature : Added Customised Error logs [Fixes : #1122]

### LINKED ISSUE #1122: No appropriate message on startup when DB is offline
**Describe the bug**
If you start Talawa API and the DB isn't started you get the message in the screen shot.

If you look carefully you will see a:
1. TCP Connection refused (ECONNREFUSED) message
2. Session timeout message

Then there is a stack trace.

This  is not a good UX. A typical non-IT administrator will not know how to proceed. 

**To Reproduce**

- See above

**Expected behavior**

- A very clear message that the DB is unreachable and that it may be offline without the stack trace.

**Actual behavior**

- See above

**Screenshots**

image

**Additional details**

aviraldevv:
Improved and customised the logs for Talawa-API

**Issue Number:**
Fixes #1122 

**Snapshots/Videos:**

Screenshot_353

Screenshot_348

**Summary**

The motivation for this change is to improve the user experience when starting Talawa API and the DB isn't started. Currently, the error message can be confusing and overwhelming for non-IT administrators. The pull request solves this problem by providing a more user-friendly error message with clear instructions on what to do next. This change will make the API more accessible to a wider audience.

Yes


aviraldevv:
@palisadoes  I have used just one `and` if you want I can rephrase it according to you.

src/db.ts
```diff
@@ -10,8 +10,27 @@ export const connect = async () => {
       useFindAndModify: false,
       useNewUrlParser: true,
     });
-  } catch (error) {
-    logger.error("Error while connecting to mongo database", error);
+  } catch (error: any) {
+    const errorMessage = error.toString();
+    if (errorMessage.includes("ECONNREFUSED")) {
+      logger.error("\n\n\n\x1b[1m\x1b[31m%s\x1b[0m", error);
+      console.log(
```
xoldyckk, line 17: Logger is used to avoid using console logs. You're console logging inside the logger itself.

aviraldevv: @xoldyckk I understand your point but I have used logger to log error and for rest of fixes I used console.log because it was looking better. I can restore it if you want.

xoldyckk: @aviraldevv You should opt for a solution that fixes the message formatting of logger itself instead of opting to use console log to make it easier for contributors to easily display that something went wrong with mongodb.

xoldyckk: @aviraldevv Yes. The task you're given is format those mongoose errors to be easily readable, maybe even add some custom stuff. But you have to do that with logger itself. You can't use console.log.

aviraldevv: @xoldyckk Please check slack.

xoldyckk:
@aviraldevv can u post a screenshot of what these new logs look like? If they're the same ones you sent me on slack then lemme know.

aviraldevv:
@xoldyckk I have updated the PR screenshot.

xoldyckk:
> @xoldyckk I have updated the screenshot in the PR. Does it look good to you??

Good enough

xoldyckk (approved):

palisadoes (approved):

PR #1134, opened 2023-03-07
Add dynamic base url support for images to work in production environments [Fixes #1131]

### LINKED ISSUE #1131: Hardcoded BASE_URL is used instead of dynamic BASE_URL in ```image.ts```
**Describe the bug**
A hardcoded base URL with ```http://localhost:${process.env.port || 4000}/``` is being used for fetching of images. Due to this, the organization images would not be fetched if Talawa Admin is hosted on any service.

**To Reproduce**

1. Go to https://talawa.app
2. Login using email ```testuser1@gmail.com``` and password ```testuser1```
3. Try creating an org with an image
4. Image will not work

**Expected behavior**
The dynamic url should be fetched from the API using req parameter of ```context``` object of GraphQL resolver.

**Actual behavior**
A hardcoded localhost:4000 url is being used everywhere.

**Screenshots**
(Not working URL)
image
.
.
Using dynamic url, it would have been (Working URL)
image

**Additional details**
This error is restricting the use of images to local environment only.
An application must be developed keeping Production Environment in mind. Otherwise, switching to production environments from local environ all of a sudden in future may require developer upfront.

anshgoyalevil:
This PR adds the dynamic apiRootUrl into context of all requests which returns a url based on the current server's hostname and protocol.

**Issue Number:**

Fixes #1131 

Yes

**Snapshots/Videos:**
Not Relevant

Not Required

**Summary**
This PR adds the dynamic apiRootUrl into context of all requests which returns a url based on the current server's hostname and protocol.

No

**Other information**
This makes the current Talawa-API images logic to work in remote production or development servers.

Yes

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -102,6 +102,15 @@ const apolloServer = new ApolloServer({
     role: RoleAuthorizationDirective,
   },
   context: ({ req, res, connection }) => {
+    /**
+     * The apiRootUrl for serving static files.
+     * This is constructed by extracting the protocol and host information from the `req` object.
+     * It is passed to the context object and can be accessed by all the resolver functions.
+     * Resolver functions can use this to construct absolute URLs for serving static files.
+     * For example, http://testDomain.com/ is apiRootUrl for a server with testDomain.com as Domain Name
+     * with no SSL certificate (http://)
+     * In local environment, apiRootUrl will be http://localhost:{port}/
+     */
     const apiRootUrl = `${req.protocol}://${req.get("host")}/`;
     if (connection) {
       return {
--- a/src/server.ts
+++ b/src/server.ts
@@ -102,12 +102,14 @@ const apolloServer = new ApolloServer({
     role: RoleAuthorizationDirective,
   },
   context: ({ req, res, connection }) => {
+    const apiRootUrl = `${req.protocol}://${req.get("host")}/`;
     if (connection) {
       return {
         ...connection,
         pubsub,
         res,
         req,
+        apiRootUrl,
       };
     } else {
       return {
@@ -115,6 +117,7 @@ const apolloServer = new ApolloServer({
         pubsub,
         res,
         req,
+        apiRootUrl,
       };
     }
   },
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -212,8 +212,6 @@ export const EMAIL_ALREADY_EXISTS_CODE = "email.alreadyExists";
 export const EMAIL_ALREADY_EXISTS_MESSAGE = "email.alreadyExists";
 export const EMAIL_ALREADY_EXISTS_PARAM = "email";
 
-export const BASE_URL = `http://localhost:${process.env.port || 4000}/`;
-
 export const ACCESS_TOKEN_SECRET = process.env.ACCESS_TOKEN_SECRET;
 
 export const REFRESH_TOKEN_SECRET = process.env.REFRESH_TOKEN_SECRET;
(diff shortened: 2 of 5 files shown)


xoldyckk:
@anshgoyalevil Should we really work hard to improve such nitty gritty functionalities for an implmentation that is temporary? Isn't this base64 encode thing going to be replaced?

Even if it weren't this isn't really the best solution. If image addresses are stored relative to api's directory instead of some external url(s3 for example), then it's the job of clients to correctly fetch it by appending that address with whatever the api's address is. The api itself cannot make assumptions on what protocol, host, port etc., the developer will make it run on.

anshgoyalevil:
@xoldyckk 

This solution has nothing to do with the base64 encoding. Even if it gets replaced, the logic for fetching images would still remain the same.

If we store files on the FS of API, we would still need to reconstruct the URL of the API server to fetch those images. No matter how images are stored in FS, they would still be fetched using the relative URLs.

Storing them using the absolute URL in database imposes a risk in case we shift the server from one service to another. In that case, the proposed solution will construct the URLs based on the current DNS.

xoldyckk:
> @xoldyckk
> 
> This solution has nothing to do with the base64 encoding. Even if it gets replaced, the logic for fetching images would still remain the same.
> 
> If we store files on the FS of API, we would still need to reconstruct the URL of the API server to fetch those images. No matter how images are stored in FS, they would still be fetched using the relative URLs.
> 
> Storing them using the absolute URL in database imposes a risk in case we shift the server from one service to another. In that case, the proposed solution will construct the URLs based on the current DNS.
> 
> In case, we are relying on some external storage bucket, S3 for example, this solution will not work. But, according to Peter, we need a solution to store files on FS as well. In that case, we can latch the server to use the other solution.

OH. So, FS will be there. Then it's ok.

But this implementation is kinda dirty, you don't need to make a new function just to get information on the running server. If it's available in the context object, use it directly inside resolver. You, can include the url information as a variable inside context object for use across multiple resolvers.

anshgoyalevil:
@xoldyckk 

Okay, I will remove the use of external function requestUrl.ts, and would just integrate that login into image.ts resolver itself.

xoldyckk:
> @xoldyckk
> 
> Okay, I will remove the use of external function requestUrl.ts, and would just integrate that logic into image.ts resolver itself.

No, do it inside context creator function. Parse/format the values from request object to create some variable like `apiRootUrl`, and pass it through context object. That way whichever resolver is making use of media files to provide absolute addresses can make use of the same variable instead of calculating it from scratch in each resolver.

anshgoyalevil:
> 

Okay. Will try figuring out a way of doing that.

anshgoyalevil:
@xoldyckk 
Made the changes.
Please review.

src/server.ts
```diff
@@ -102,19 +102,22 @@ const apolloServer = new ApolloServer({
     role: RoleAuthorizationDirective,
   },
   context: ({ req, res, connection }) => {
+    const apiRootUrl = `${req.protocol}://${req.get("host")}/`;
```
xoldyckk, line 105: Add a detailed comment here why this is needed.

anshgoyalevil: @xoldyckk Done

xoldyckk:
> @xoldyckk Made the changes. Please review.

Those are some good comments. Nice job.

xoldyckk (approved):

palisadoes (approved):

PR #1158, opened 2023-03-10
Implement the functionality to create tags for user

### LINKED ISSUE #1126: Feature Request: Implement tags for User in an organization
As per the discussion in [this thread](https://github.com/PalisadoesFoundation/talawa-admin/issues/496), we must work on implementing support for `tags` on the `User` model as a part of an `Organization`.

1. Create a new model called `UserOrgTags` which will have the following fields:
- `User`: Stores the userID
- `OrganizationID`: Stores the organization
- `tags`: Stores the array of tags that a user has been assigned for an organization

2. Create a new field called tags on the `User` schema which accepts a `OrgID` as an input and returns all the tags that the user has in that organization.
3. Create a new field called `allTags` on the `Organization` schema which returns all the `distinct` tags that have been allotted to across all users. 

NA

Described above

EshaanAgg:
Implements all the necessary models and mutations to enable users to be categorised by different tags. 

- Each tag is treated as a `folder` which can house other tags as well.
- Each tag is described by its `title`, and the tags of an user depend upon the organization.
- While only the admins of an organization and `SUPERADMIN` can create, update, delete, assign and unassign tags, the tags can be viewed publically by all of the users without any authentication.
- Tags of a user are directly exposed on their Model itself, and can be accessed by providing the `organizationId` of the organization of whose tags you want to view.

**Issue Number:**

Fixes #1126 

In the process. Added tests for all resolvers of the `Tag` model.

**Snapshots/Videos:**
NA

NA

**Summary**
Implementing the tagging feature for sorting of users, as implemented by Breeze CHMS.

No

**Other information**
NA

Yes


jadeharr:
@EshaanAgg The code coverage from the Codecov report has decreased. Please fix your code so the level of coverage is at least maintained.

xoldyckk:
@EshaanAgg I told you to not commit the resolvers yet. Now everyone will complain about coverage lol.

EshaanAgg:
Lol just saw your comment right now. Can you please leave a changes requested review so that people can know I'm working on it? Meanwhile you can review the models and the schema.

src/typeDefs/mutations.ts
```diff
@@ -67,6 +69,8 @@ export const mutations = gql`
 
     createPost(data: PostInput!, file: String): Post @auth
 
+    createTag(title: String!, parentTag: ID, organizationId: ID): Tag @auth
```
xoldyckk, line 72: Same as `toggleTagAssign`, make the input type, follow the naming convention.
Rename `title` to `name`, `parentTag` to `parentTagId`.

src/typeDefs/queries.ts
```diff
@@ -108,6 +108,8 @@ export const queries = gql`
 
     users(where: UserWhereInput, orderBy: UserOrderByInput): [User] @auth
 
+    usersByTag(tagId: ID!): [User] @auth
```
xoldyckk, line 111: Remove this, this can easily be queried using `tag.usersAssignedTo` relation connection.

src/models/UserTag.ts
```diff
@@ -0,0 +1,31 @@
...
+  tag: PopulatedDoc<Interface_Tag & Document>;
+}
+
+const UserTagSchema = new Schema({
+  user: {
+    type: Schema.Types.ObjectId,
+    ref: "User",
+    required: true,
+  },
+  tag: {
```
xoldyckk, line 17: Rename to `tagId`

xoldyckk (commented):
Make these changes. After that we'll see what else can be improved. If you think something I'm asking would be bad pattern lemme know.

src/models/Tag.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+];
+
+const TagSchema = new Schema({
+  title: {
+    type: String,
+    required: true,
+    validate: validateTagName,
+  },
+  organization: {
```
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/models/Tag.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+  parentTag: {
+    type: Schema.Types.ObjectId,
+    ref: "Tag",
+    required: false,
+    default: null, // A null parent corresponds to a root tag in the organization
+  },
+});
+
+TagSchema.index({ organization: 1, parentTag: 1 });
```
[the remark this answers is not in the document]
EshaanAgg: Got it. Great catch.

src/models/Tag.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Interface_Organization } from "./Organization";
+
+export interface Interface_Tag {
+  _id: Types.ObjectId;
+  organization: PopulatedDoc<Interface_Organization & Document>;
+  parentTag: PopulatedDoc<Interface_Tag & Document>;
+  title: string;
+}
+
+const validateTagName = [
```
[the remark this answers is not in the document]
EshaanAgg: Okay. Will remove these validations as of now.

src/models/UserTag.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Interface_User } from "./User";
+import { Interface_Tag } from "./Tag";
+
+export interface Interface_UserTag {
+  _id: Types.ObjectId;
+  user: PopulatedDoc<Interface_User & Document>;
+  tag: PopulatedDoc<Interface_Tag & Document>;
+}
+
+const UserTagSchema = new Schema({
```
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/typeDefs/types.ts
```diff
@@ -294,6 +295,15 @@ export const types = gql`
     aggregate: AggregatePost!
   }
 
+  type Tag {
+    _id: ID!
+    title: String!
+    organization: Organization!
+    parentTag: Tag
+    childTags: [Tag]
```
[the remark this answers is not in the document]
EshaanAgg: I am providing support for these as the front-end might need to display the hierarchy of tags. So in a certain view, they might want to query for the parent tag of the current tag, or the child ones. 
Providing support for the same is a good pattern in my opinion, as we are not storing any additional data for the same. They would be implemented using field-resolvers, thus would only run when specially queried for.

src/typeDefs/types.ts
```diff
@@ -294,6 +295,15 @@ export const types = gql`
     aggregate: AggregatePost!
   }
 
+  type Tag {
+    _id: ID!
+    title: String!
+    organization: Organization!
+    parentTag: Tag
```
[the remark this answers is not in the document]
EshaanAgg: Same logic as described in comment above.

src/typeDefs/types.ts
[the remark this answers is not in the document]
xoldyckk: Hmm, unless you're building some kind of sophisticated UI design I don't see where this'll be used. The tags for an organization will just be fetched as a list, then the UI will be rendered according to some function logic which will render each tag successively according to parentTagId and the UI behaviour will also depend on parentId(for example adding padding to the left of tags, for each successive child tag).

I personally think it'll be never used, and it's a very niche feature, but if you want to provide this functionality then sure.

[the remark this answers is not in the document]
xoldyckk: Nevermind, you're correct. If the tags are to be lazily loaded on the client side, this will be a required. But this feature can only truly be used when there's also a query field for a tag or a node interface query field.

I just thought maybe returning all tags in organization's tags field is a bad idea. We should only return the tags that have `parentId === null` for that organization. Child tags should always be lazily loaded. What do you think?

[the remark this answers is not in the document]
EshaanAgg: Okay got your concerns!

src/typeDefs/types.ts
```diff
@@ -199,6 +199,7 @@ export const types = gql`
     apiUrl: URL!
     createdAt: DateTime
     pinnedPosts: [Post]
+    rootTags: [Tag]
```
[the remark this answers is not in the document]
EshaanAgg: Got it. Will make `tags` a graphQL connection called `TagConnection`. But then we would also have to introduce a query right, called `organizationTagsConnection`?

PS. This is my first time working with connections, so these forgive me if this seems like a silly doubt!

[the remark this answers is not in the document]
xoldyckk: @EshaanAgg This'll require that you either implement a node interface query field or a seperate tag query field. Either one's fine, the node interface one is better for relay graphql based clients.

Ulimately, the main thing is that related entities and connections should be present as relations, and not as query fields. Query fields are reserved for objects which must be refetchable by their unique ids or connections which are not directly related to their parents. 

For example in instagram, posts of a user are directly related to that user, while posts in the explore section are not. So, the posts in explore section must have their own connection query field like while posts for a user can just be queried using `user.postsConnection` relation.

src/resolvers/Mutation/assignTag.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  if (!tag) {
+    throw new errors.NotFoundError(
+      TAG_NOT_FOUND.MESSAGE,
+      TAG_NOT_FOUND.CODE,
+      TAG_NOT_FOUND.PARAM
+    );
+  }
+  // Check that the user shouldn't already be assigned the tag
+  const userAlreadyHasTag = await UserTag.exists({
```
beingnoble03, line 45: @EshaanAgg @xoldyckk I have a doubt here. As we are only dealing with the tags (no tagFolders anymore), so, what if we pass the child TagId here, we'll assign the child tagId here, even if he's already assigned the parent tagId.
To resolve this issue, we have 2 approaches that we can follow here
- Either we assign user all the child Tags too, when any parent Tag is assigned.
- Either we add a check here, to check if the tagId belongs to any child TagId of the previously assigned parent tagId.
What should be done?

beingnoble03: I think this was implemented for the previously thought approach of `tagFolder`.

xoldyckk: @beingnoble03 I haven't reviewed anything except mongoose models and graphql typeDefs. Let's first have a proper strucuture to work with.

Regarding your question I'm not sure myself. It'll be left as a decision for @palisadoes.

In gmail, if you assign something a parent label, the parent label along with all its child labels, no matter how deeply nested are also applied to the email. I explained logic for this over [here](https://github.com/PalisadoesFoundation/talawa-api/issues/1126#issuecomment-1462372110).

If @palisadoes wants we can put together a logic to only apply labels if they have no successive child. The logic will work like this:-

```
school

school/class1/mathematics
school/class1/science

school/class2
```

So, if you assign someone a label of `school` or `school/class1`, it'll fail. It'll only work for `school/class1/mathematics` and `school/class1/science`. It'll also work for `school/class2` as it has no successive child.

palisadoes:
I was looking at the TypeDefs. The design limits us to tags for people. How can it be made more flexible so that we can use them for any resource without constantly creating new structures? For example how could these be dynamically added?:

1. People
2. Events
3. Venues
4. Skills
5. Characteristics
6. etc

What we have is limiting.

src/resolvers/Mutation/assignTag.ts
EshaanAgg: It makes more sense to follow the Gmail like logic in this case. 

As from the implementation point of view, any user would just be assigned their immediate parent in the database. Then when queries with a tag, we can use graph lookup to first query all of the child tags, and then make another query to fetch all the users belonging to any of those parent or child tags.

A similar logic could be implemented for users as well, but I feel from the frontend perspective, it would make the most sense to return the deepest tags only (as we might to show 'married' tag on the user's profile, but showing the 'marital status' tag makes no sense). However the same could be queried with the help of the current fields if the need arises.

EshaanAgg:
The current design uses Tags as something that can be attached to an `Organization` and then we are just using the `TagUser` schema to assign the tags to some users. If we wish to assign tags to `Events` or other objects, lets say like `Posts`, the same can be done by just creating a new schema.

The current implementation would require the resolvers to be rewritten in such a case. To make this assigning of tags `model-agnostic` (if you would allow that), we would need to migrate towards the relay specification and implement a `Node` interface to work with all these objects. Implementation of the same is a complete redo of out GraphQL design for the most part as I see it, but I'm not as well versed with `GraphQL` as @xoldyckk!

He would be able to better guide us.

EshaanAgg:
@xoldyckk Updated the `TypeDefs` as per your comments. What do you think?

src/resolvers/Mutation/assignTag.ts
xoldyckk: @EshaanAgg 

If you're talking about how tags will work in talawa-admin for filtering and stuff, they'll work how the image for breeze cms shows us. First you fetch list of parent tags(it'll be paginated using graphql connection), when you click on the tag you see all users who are related to that tag. There'll be an **_plus(+)_** button on the tag, when you click it, it'll load its child tags. Clicking on each child tag will reveal users related to that tag. This approach will be followed for however deeply nested the tags are, until you reach the last nested tag level. Like this:-

_**school +**_

_**school -**_
&nbsp;&nbsp; _**class1 +**_
&nbsp;&nbsp;**_class2_ +**

_**school -**_
&nbsp;&nbsp; _**class1 -**_
&nbsp;&nbsp;&nbsp;&nbsp;_**science**_
&nbsp;&nbsp;&nbsp;&nbsp; _**mathematics**_
&nbsp;&nbsp;**_class2_ -**
&nbsp;&nbsp;&nbsp;&nbsp; _**science**_
&nbsp;&nbsp;&nbsp;&nbsp; **_mathematics_**

Each subsequent nested level will only be loaded when you click on that **_+_** icon. When you click on **_-_** it'll be minimized.

Regarding what you said about only sending the deepest tags, I take it you meant how to display it on a user's profile/information. I don't think that's good. For example you're viewing a user's profile, for a particular organization. The child tag by itself will make no sense.

What you're suggesting:-

**_science_**

This is what I think it should be:-

**_school/class1/science_**

EshaanAgg: Yes @xoldyckk. For the display on the navigation panel, I was thinking the same. The description I had provided earlier was more related to how to how the assigning of roles would work. Like if an event I want to add `class2` as attendees, then to implement the same our resolver would make the two database queries, one to get the complete subtree of the the parent tag (that is `class2`, `class2/science`, `class2/mathematics`), and then other query to fetch the details of all the users assigned to any of these tags.

As for the display, the primary reason I was advocating for the earlier approach  was that as a user starts to be allotted multiple tags, the above way of display gets too cluttered too fast. Anyhow we can decide on the same later, once the definitions and structures are finalised.

xoldyckk:
> I was looking at the TypeDefs. The design limits us to tags for people. How can it be made more flexible so that we can use them for any resource without constantly creating new structures? For example how could these be dynamically added?:
> 
> 1. People
> 2. Events
> 3. Venues
> 4. Skills
> 5. Characteristics
> 6. etc
> 
> What we have is limiting.

Each tag has to be uniquely connected to another object be it user or something else like an event, venue etc. So, for each many to many relation between the tag and <OBJECT> we will need a new junction collection(table). 

It'll be like this:-
```
tagId userId
1         1          
```
```
tagId eventId
1         1    
```
```
tagId venueId
1         1    
```
This is that this approach is better if you have more related information you want to store specifically between the tag and the partnering object. This approach is cleaner,easier to understand and implement. The problem with this approach is that it will take more database space. 

Of course this can also be accomplished by cramming all of those objects within a table like `TagObject` where object can be user, event, venue etc. But it'll be a very dirty imlementation. At any time when a tag is connected to an object, foreign keys(objectIds) of all other objects will need to be null. In a single row(field) of `TagObject` collection(table), the tag can only be related to one of those objects.

It'll be like this:-
```
tagId   userId    eventId     venueId        ...more objects here
1         1         null        null         ...more nulls here
1         null      1           null         ...more nulls here
1         null      null       1             ...more nulls here
```
This approach will require more custom logic to check those nulls. It's dirty yes, but it's more database efficient approach for sure.

src/resolvers/Mutation/assignTag.ts
xoldyckk: @EshaanAgg At anytime, the tag's id will be available wherever the tag is used. You only need the tagId to query the users related to it.

I didn't understand what you said, explain with some diagrams man, like I'm making these badass tables.

EshaanAgg: Ahhh I wish I was as badass as you, so I'll just copy yours!

For the given structure:
```
school
   class1 
     science
     mathematics
  class2 
     science
     mathematics
```

Let's say we are working with events (PTM for example), and we want to assign all the `class2` students to be attendees of the same. So in such a case, with the current implementation, the frontend would make to make a singular query of the sort `allAttendee(tag: class2, event: PTM)`. 

On the backend, we would receive the given query and then in effect make two database queries. The first one would be a `graphLookup` query of the `tags` model, that would return the subtree of the given input tag.
In our case, it would be the tags `school/class2`, `school/class2/science` and `school/class2/mathematics`. 

Then we would make a second query, this time to the `TagUser` model, to delete all the entries if their `tag` belongs to the list of the tags generated earlier. A similar approach will also be used in the `removeTag` mutation.

src/typeDefs/types.ts
```diff
@@ -332,6 +342,7 @@ export const types = gql`
     pluginCreationAllowed: Boolean
     adminApproved: Boolean
     createdAt: DateTime
+    tags(organizationId: ID!): [Tag]
```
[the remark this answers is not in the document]
xoldyckk: @EshaanAgg Why'd you remove this one?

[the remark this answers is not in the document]
EshaanAgg: Added it as a graphQL connection query.

src/resolvers/Mutation/assignTag.ts
xoldyckk: @EshaanAgg

First of all frontend needs to make a query with tagId not tagName. What they see on their screen will be tagName, but what gets sent in mutation will be tagId. TagId can be uniquely identified in `TagUser` junction, because we'll have the tagId.

IDK what you mean when you say we need to fetch subtree of tags. We can get all users related to a tag whether it has nested tags or not in a single query. Look at [this](https://github.com/PalisadoesFoundation/talawa-api/issues/1126#issuecomment-1462372110) again. When I say they'll get assigned to parent tags, I mean in `TagUser` table the user will get a new row(document) connected to each of those tags. If a tag has `n` levels of nesting with 2 items for each nesting level, the number of `TagUser` rows(documents)  created will be `2^n - 1`.

If you want to add all users with tag `school/class1` to the event `PTM`, just make the mutation with `school/class1` tag's id.

This is how gmail's lablels works, I know this is too much duplication of data and it can be avoided, but that'll require another complex implementation. This also will perform worse during mutations, the more levels of nesting you have, that's why I asked whether the levels of nesting should be finite. But @palisadoes said it has to be infinite so, I guess not.

src/typeDefs/types.ts
```diff
@@ -10,6 +10,10 @@ export const types = gql`
     count: Int!
   }
 
+  type AggregateTag {
```
xoldyckk, line 13: This type is useless tbh. We can directly expose the tagsCount on User and Organization types, or directly inside their nested tags connection.

palisadoes:
> > I was looking at the TypeDefs. The design limits us to tags for people. How can it be made more flexible so that we can use them for any resource without constantly creating new structures? For example how could these be dynamically added?:
> > 
> > 1. People
> > 2. Events
> > 3. Venues
> > 4. Skills
> > 5. Characteristics
> > 6. etc
> > 
> > What we have is limiting.
> 
> Each tag has to be uniquely connected to another object be it user or something else like an event, venue etc. So, for each many to many relation between the tag and we will need a new junction collection(table).
> 
> It'll be like this:-
> 
> ```
> tagId userId
> 1         1          
> ```
> 
> ```
> tagId eventId
> 1         1    
> ```
> 
> ```
> tagId venueId
> 1         1    
> ```
> 
> This approach is better if you have more related information you want to store specifically between the tag and the partnering object. This approach is cleaner,easier to understand and implement. The problem with this approach is that it will take more database space.
> 
> Of course this can also be accomplished by cramming all of those objects within a table like `TagObject` where object can be user, event, venue etc. But it'll be a very dirty imlementation. At any time when a tag is connected to an object, foreign keys(objectIds) of all other objects will need to be null. In a single row(field) of `TagObject` collection(table), the tag can only be related to one of those objects.
> 
> It'll be like this:-
> 
> ```
> tagId   userId    eventId     venueId        ...more objects here
> 1         1         null        null         ...more nulls here
> 1         null      1           null         ...more nulls here
> 1         null      null        1             ...more nulls here
> ```
> 
> This approach will require more custom logic to check those nulls. It's dirty yes, but it's more database efficient approach for sure.
> 
> One more approach is this:-
> 
> ```
> tagId   objectId   object_type
> 1         1           user
> 1         1           event
> 1         1           venue
> ```
> 
> But this one's performance will degrade as the number of entries increases.

@xoldyckk @EshaanAgg @beingnoble03 as the the three participants here

1. Yes, there could be a performance issue, but this means the user will not get what they want if we extend the functionality. This then implies that users will not get what they want if we cannot optimize to perfection. It limits the attractiveness of the application.
2. The amount of space tags will add as a percentage of the  overall  the environment  (database and/or cloud storage) will be tiny. I strongly suggest that you think of a way to make this extensible.
3. It appears that GitHub for example uses this approach, issues, PRs and discussions all get sequentially generated IDs for the repo. Try opening an issue and immediately creating a PR for example and you will see that it is sequential.
4. Do not limit this solution to users from the start, we will create a serious long term problem.

Figure out how to do this as efficiently as possible.

DMills27:
I think you guys are overcomplicating it. Here's what I'd do if I had to create tags from scratch.

Define two main GraphQL types: `User` and `Tag`. The User type would contain fields like `id`, `name`, `email`, and `organizationId`, along with a list of tags that the user belongs to. The `Tag` type would contain fields like `id`, `title`, `parentTagId` (if it's a nested tag), and `organizationId`.

Here's an example schema for these types:

```GraphQL
type User {
  id: ID!
  name: String!
  email: String!
  organizationId: ID!
  tags: [Tag!]!
}

type Tag {
  id: ID!
  title: String!
  parentTagId: ID
  organizationId: ID!
}

type Query {
  getUserTags(organizationId: ID!): [Tag!]!
}

type Mutation {
  createTag(title: String!, organizationId: ID!, parentTagId: ID): Tag!
  updateTag(id: ID!, title: String!, parentTagId: ID): Tag!
  deleteTag(id: ID!): ID!
  assignTag(userId: ID!, tagId: ID!): User!
  unassignTag(userId: ID!, tagId: ID!): User!
}
```

The `getUserTags` query allows users to view the tags of an organization without authentication. The `createTag`, `updateTag`, `deleteTag`, `assignTag`, and `unassignTag` mutations are protected by authentication and only admins of an organization or SUPERADMINs can access them.

To enable tags to be nested within other tags, you could use a self-referencing `parentTagId` field in the `Tag` type. This would allow you to create a tree-like structure of tags, where each tag can have child tags.

The `assignTag` and `unassignTag` mutations would allow admins to assign or unassign tags to users, updating the tags field on the User type accordingly.

Overall, this schema would enable users to be categorized by tags and allow admins to manage those tags for their organization.

xoldyckk:
> I think you guys are overcomplicating it. Here's what I'd do if I had to create tags from scratch.
> 
> Define two main GraphQL types: `User` and `Tag`. The User type would contain fields like `id`, `name`, `email`, and `organizationId`, along with a list of tags that the user belongs to. The `Tag` type would contain fields like `id`, `title`, `parentTagId` (if it's a nested tag), and `organizationId`.
> 
> Here's an example schema for these types:
> 
> ```graphql
> type User {
>   id: ID!
>   name: String!
>   email: String!
>   organizationId: ID!
>   tags: [Tag!]!
> }
> 
> type Tag {
>   id: ID!
>   title: String!
>   parentTagId: ID
>   organizationId: ID!
> }
> 
> type Query {
>   getUserTags(organizationId: ID!): [Tag!]!
> }
> 
> type Mutation {
>   createTag(title: String!, organizationId: ID!, parentTagId: ID): Tag!
>   updateTag(id: ID!, title: String!, parentTagId: ID): Tag!
>   deleteTag(id: ID!): ID!
>   assignTag(userId: ID!, tagId: ID!): User!
>   unassignTag(userId: ID!, tagId: ID!): User!
> }
> ```
> 
> The `getUserTags` query allows users to view the tags of an organization without authentication. The `createTag`, `updateTag`, `deleteTag`, `assignTag`, and `unassignTag` mutations are protected by authentication and only admins of an organization or SUPERADMINs can access them.
> 
> To enable tags to be nested within other tags, you could use a self-referencing `parentTagId` field in the `Tag` type. This would allow you to create a tree-like structure of tags, where each tag can have child tags.
> 
> The `assignTag` and `unassignTag` mutations would allow admins to assign or unassign tags to users, updating the tags field on the User type accordingly.
> 
> Overall, this schema would enable users to be categorized by tags and allow admins to manage those tags for their organization.

All this is already done. What we're discussing is how the data is modeled in database and how it will be mutated.

xoldyckk:
@DMills27 All this is already done. What we're discussing is how the data is modeled in database and how it will be mutated.

src/resolvers/Mutation/assignTag.ts
xoldyckk: @EshaanAgg I just tried gmail myself and what I read on google about gmail's labels was wrong. Gmail doesn't automatically assign child labels of a parent label to the email which currently has the parent label assigned to it. So, [this](https://github.com/PalisadoesFoundation/talawa-api/issues/1126#issuecomment-1462372110) explanation of mine is incorrect.

src/typeDefs/mutations.ts
```diff
@@ -164,6 +170,8 @@ export const mutations = gql`
 
     updatePluginStatus(id: ID!, status: Boolean!): Plugin!
 
+    updateTag(tagId: ID!, newName: String!): Tag @auth
```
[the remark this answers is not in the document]
EshaanAgg: Ahh got it. Thanks!

src/typeDefs/queries.ts
```diff
@@ -102,6 +102,15 @@ export const queries = gql`
 
     tasksByUser(id: ID!, orderBy: TaskOrderByInput): [Task]
 
+    tagsAssignedWithByUserConnection(
```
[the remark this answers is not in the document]
EshaanAgg: Ahh got it. Shouldn't we also create a custom input type here for the arguments of tagsAssignedWith?

DMills27:
Sorry, I miss understood. Here's how I'd model it in the database if that's the case. 

Database Model:

- Store the tags as documents in a `tags` collection with fields for `id`, `title`, `organization_id`, and `parent_tag_id`.
- Use a tree structure to enable nesting of tags, with each tag document containing a reference to its parent tag as an embedded document or a reference to the parent tag document ID.
- Store the user tags as documents in a `user_tags` collection with fields for `user_id` and an array of `tag_ids`, indicating which tags are assigned to which users.

Assigning Tags to Users:

- To assign a tag to a user, you could update the corresponding user document in the `user_tags` collection by adding the tag ID to the `tag_ids` array field.

- To unassign a tag from a user, you could update the corresponding user document in the `user_tags` collection by removing the tag ID from the `tag_ids` array field.

Client-side Usage:

- To retrieve the list of tags for an organization, you could use a MongoDB query to fetch the tags documents with a filter for the `organization_id` field. This query could be publicly available to all users without any authentication.

- To perform mutations on tags, such as creating, updating, or deleting a tag, you would need to use a MongoDB query or API to create, update, or delete the corresponding `tags` document. This would require appropriate authentication and authorization to ensure that only authorized users can perform these operations.

- To assign or unassign tags to users, you could use a MongoDB query or API to update the corresponding `user_tags` document, adding or removing the appropriate `tag_id` from the `tag_ids` array field.

- To update a tag for a user, you could use a combination of the create and delete operations. First, remove the existing tag from the `tag_ids` array field, then add the updated tag to the `tag_ids` array field.

- On the client-side, you could use the assigned tags to filter and display users, or perform other operations based on the tags assigned to each user.

Overall, this approach uses a document-oriented database model to store tags and assign them to users, and allows mutations to be performed based on those tags using MongoDB queries or APIs. It also enables granular access control, allowing only authorized users to perform certain mutations, while keeping the tags themselves publicly visible to all users.

src/typeDefs/queries.ts
```diff
@@ -102,6 +102,15 @@ export const queries = gql`
...
+      userId: ID!
+      orgId: ID!
+      first: Int
+      skip: Int
+    ): TagConnection
+
+    tagsByOrganizationConnection(id: ID!, first: Int, skip: Int): TagConnection
```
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/typeDefs/types.ts
EshaanAgg: Got it. Will remove the same.

src/typeDefs/mutations.ts
```diff
@@ -67,6 +67,8 @@ export const mutations = gql`
 
     createPost(data: PostInput!, file: String): Post @auth
 
+    createTag(input: createTagInput): Tag @auth
```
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/typeDefs/queries.ts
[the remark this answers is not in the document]
xoldyckk: @EshaanAgg For connections no. I know it makes sense to use input types. But I've read about talks on it that it adds more overhead, idk what graphql guys mean by overhead here, but they say that the more complex input object you have, the slower that graphql request will be.

Also, it's the standard in all graphql ecosystem to provide connection arguments like this, according to relay server specification so you just have to follow it. All tools and libraries generated around graphql expect connections to have these arguments, defined exactly like this.

xoldyckk:
@DMills27 
Please check out the changes first, all this is already implemented. We're talking about having a flexible approach for tags here.

Also we're using junction collections here instead of just embedding objectIds as lists, as number of tags can be extremely large and also because this approach is not flexible. It becomes slower the more objectIds you have inside that list, and it's very hard to make complex aggregate/join database queries etc with it.

src/resolvers/Mutation/assignTag.ts
EshaanAgg: @xoldyckk 
1. I know that and would use the `_id` properties of the tag only to work with them (on frontend) as obviously the name's of the tags would not be unique across the different folders, and would be bad practice in general. The choice of using names in the examples was just done for the ease of implementation.

2. Ahh got it. I was advocating for the other approach, where on assigning the tag `school/class2/science` to `userA` would only create a single row in the `TagUser` schema, corresponding to 
```
userA    school/class2/science
```
and not 
```
userA    school
userA    school/class2
userA    school/class2/science
```

The feel the later approach is bad, not because its duplication of data, but it is exponential duplication of data. The whole point of this flexibility is to allow users to nest the tags as deep as they want, and when we keep this consideration in mind, the amount of overhead generated is concerning.

With my approach, although the resolver logic would be difficult to implement, the same would achieve the desired effect with less overall overhead. For example, if we query to get all the users with the tag `class2`, we will first traverse the `subtree` of tags of `class2`, that is all the tags that have this tag as their ancestor (immediate or other). In our case they are `school/class2/science` and `school/class2/mathematics`.
Not we can perform the required operation in the `UserTag` scheme, but instead of using an equality in id, i.e. 
```ts
UserTag.find({
   _id : args.id,
})
```

we will first generate this list of child tags, and then use the `$in` clause:
```ts
const allChildTags = []     // contains school/class2, school/class2/science and school/class2/mathematics 
UserTag.find({
   _id : {
     $in: allChildTags,
    }
})
```

3. Oops. What do you think we should continue with now? Should we follow the gmail based approach, or continue with this `child also gets parent` tag approach.

xoldyckk: @EshaanAgg Gmail's approach is better. We were scratching our heads uselessly because we thought gmail used some complex checks in their backend or something to make that previous approach work. But their implementation is very simple. Read the link I posted above. What we were doing was premature optimization.

EshaanAgg: Got it. It will now be a simple one-to-one relationship, which would put the burden on client, and not on us.

src/typeDefs/queries.ts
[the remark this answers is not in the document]
EshaanAgg: Ahh alright got it! Pushing the updated typedefs in a couple of minutes!

EshaanAgg:
> One more approach is this:-
> 
> ```
> tagId   objectId   object_type
> 1         1           user
> 1         1           event
> 1         1           venue
> ```
> 
> But this one's performance will degrade as the number of entries increases.

I think if we are make this generalised approach of assigning tags, then this is the approach that we should be proceeding with. This is sort of the middle-ground, that is not the best performer for as data that is infinitely scalable, but it's all the simplest and the most verbose approach that we could take while providing this type of generalization.

DMills27:
@xoldyckk can you explain how your approach handles any cycles that may occur in the relationship between tags. It seems to lose the tree like structure in the way you have in mind and that in itself is a huge drawback?

xoldyckk:
> @xoldyckk can you explain how your approach handles any cycles that may occur in the relationship between tags. It seems to lose the tree like structure in the way you have in mind and that in itself is a huge drawback?

https://github.com/PalisadoesFoundation/talawa-api/pull/1158#discussion_r1133071393

palisadoes:
> > One more approach is this:-
> > ```
> > tagId   objectId   object_type
> > 1         1           user
> > 1         1           event
> > 1         1           venue
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But this one's performance will degrade as the number of entries increases.
> 
> I think if we are make this generalised approach of assigning tags, then this is the approach that we should be proceeding with. This is sort of the middle-ground, that is not the best performer for as data that is infinitely scalable, but it's all the simplest and the most verbose approach that we could take while providing this type of generalization.

This is a hard requirement. By limiting tags to users you limit creativity in how this powerful feature could be used. We will find best in class capabilities by integrating this now.

EshaanAgg:
Got it @palisadoes!

Should I work on updating the typeDefs and the models to implement this generalised approach currently for the User and Events types as a proof of concept? 
@xoldyckk

EshaanAgg:
@palisadoes @xoldyckk 
I have updated the models and the typeDefs to reflect the changes to make the implementation such that the tags can be assigned to two types: `User` and `Event`.

All the subsequent implementations for any type of other object would just involve configuring at three locations:
1. Adding the same to the enum in `src/models/TagAssign.ts -> objectType field`
2. Adding a `tagsAssignedWith` connection on the <ObjectType>. 
3. Adding a `<ObjectName>AssignedTo` field on the tag model (if we see the utility of getting all the objects of that kind assigned with a particular tag).

In my opinion, point 1 is the actual configuration step that is needed to get the same working, while steps 2 and 3 are mere support for consuming tags in the <Object> created.

beingnoble03:
@EshaanAgg @xoldyckk stated 3 approaches in [this](https://github.com/PalisadoesFoundation/talawa-api/pull/1158#issuecomment-1464892099) comment. So, are we going for the 2nd approach?

xoldyckk:
> @palisadoes @xoldyckk I have updated the models and the typeDefs to reflect the changes to make the implementation such that the tags can be assigned to two types: `User` and `Event`.
> 
> All the subsequent implementations for any type of other object would just involve configuring at three locations:
> 
> 1. Adding the same to the enum in `src/models/TagAssign.ts -> objectType field`
> 2. Adding a `tagsAssignedWith` connection on the .
> 3. Adding a `<ObjectName>AssignedTo` field on the tag model (if we see the utility of getting all the objects of that kind assigned with a particular tag).
> 
> In my opinion, point 1 is the actual configuration step that is needed to get the same working, while steps 2 and 3 are mere support for consuming tags in the created.
> 
> What do you think about this implementation?

Hold up. Prematurely making tags available for every kind of object related to an organization will introduce weird quirks later on. Think of what the tags will be used for.

Right now the tags are similar to how discord servers have tags for server members(organizations here can be compared to discord servers, and users to server members). Server members are assigned different tags in different discord servers. Those tags are also used to assign different authorities and roles to users within that server.

For talawa what's the use case for assigning these tags to events and other objects? 

If just like users, events also need to be sorted according to tags, then it's a valid use case. But there's also a complexity here. What relation does a tag for a user have to tag for an event? How will you define this in graphql schema for querying?

If events need tags just like organizations, then our current approach will not work. For that approach, we need to define a seperate tag schema which is not related to organizations, but to events themselves. Then the users can be related to tags of an event.

EshaanAgg:
As I see it, tags should be primarily be treated as objects that are used for filtering and categorisation. With the case of `User` schema, they find additional uses, such as in authorization, but that isn't the primary feature or even the use case of the same. A more apt analogy would the tags on GitHub used to classify and sort issues and PR's.

In such a case, I don't think so we will find a case of a query like `get all objects tagged with this tag`. Instead, probably we would always be certain about the object, like `all the Users tagged with this tag` or `all events tagged with this tag`. 
Thus for such a case, it makes to have different fields, like
```
type Tag {
    _id: ID!
    name: String!
    organization: Organization!
    parentTag: Tag
    childTags: [Tag]
    usersAssignedTo: [User]
    eventsAssignedTo: [Event]
  }
 ```
 
 If the need arises, we can even add a union type to get all the objects, like the following, but the same wouldn't actually be used most probably practically.
```
 type TaggedObject = User | Event

 type Tag {
    _id: ID!
    name: String!
    organization: Organization!
    parentTag: Tag
    childTags: [Tag]
    assignedTo: [TaggedObject]
 }
```
This implementation would require a `__resolveType` function to be implemented for the `assignedTo` field, which shouldn't be a major hiccup.

**Comment by xoldyckk** [comment]
@EshaanAgg So, you're saying tags are free for all within an organization, can be assigned to literally anything, and that it's the admin's job to make sure they create reasonable and sensible tags for different objects?

**Comment by palisadoes** [comment]
Some tags and folders would need to be created by default. We can figure out what those would be later

**Comment by EshaanAgg** [comment]
Got it @palisadoes! 
What do you think about the current typedefs? @xoldyckk

**Comment by xoldyckk** [comment]
@palisadoes Are you sure this is the intended feature? Grouping together related events using tags and subtags?

**Comment by EshaanAgg** [comment]
> @palisadoes 
> 
> 
> 
> Aside from grouping together users are you sure grouping together related events using tags and subtags is an intended feature?

@palisadoes Can you please clarofy the same regarding the functionality of tags.

**Comment by palisadoes** [comment]
> @palisadoes
> 
> Aside from grouping together users are you sure grouping together something like events using tags and subtags is an intended feature?

We cannot assume that the current case of only tagging users will be sufficient. Suppose an organization is a gym and somebody wants to search all the classes available in a calendar for the type they want:

1. Spin class
2. Yoga
3. Zumba
4. Aquaerobics
5. Pilates

Schools would have similar requirements. 

There may be a desire to filter for the availability of a room for an event. Now you have a location that could be included in the filtering. The possibility of accommodating resources beyond people and events will need to be included in this design.

**Comment by xoldyckk** [comment]
@EshaanAgg Make seperate junctions for different tag-object relations. Using a single junction will only create problems in the long run. The biggest problem of this approach is consistency, mongodb has no way of knowing what object is being referenced as `objectId` without the `ref` keyword. Without that it won't even check whether the object exists, before making the relation.

**Comment by xoldyckk** [comment]
@palisadoes Can you make sure that tags which are created for users can also be used to tag events? For example, in UI you'll be able to see both the user object and event object under the same tag, is this what you want?

**Comment by beingnoble03** [comment]
@xoldyckk I think while fetching the tags for the `users`, we will need to pass a variable in the query from the client side, and in the query resolver, we will need to handle the argument passed, and will return only the `users` associated with that tag. Same goes for `events`.

**Comment by xoldyckk** [comment]
Screenshot_20230314_150742

**Comment by palisadoes** [comment]
If it's a user related screen there would only need to be user tags and folders. For events it would be event tags and folders. They would be contextual.

Why do you ask?

File: src/typeDefs/types.ts [hunk 18]
```diff
@@ -294,6 +311,26 @@ export const types = gql`
...
+    _id: ID!
+    name: String!
+    organization: Organization!
+    parentTag: Tag
+    childTags: [Tag]
+    usersAssignedTo: [User]
+    eventsAssignedTo: [Event]
+  }
+
+  type TagsAssignedWithConnection {
```
xoldyckk, line 324: You don't need a different name for each connection related to same object. Just name it `TagsConnection` and use it wherever you're returning a big list of tags.

xoldyckk (commented):
@EshaanAgg You should also read a bit about mongodb indexing. Those fields which are distinct/unique/sequential etc., or important to sort/filter by (_either standalone or in combination like combination of user_id and organization_id_) should be indexed.

Think about what the requirements on talawa-admin will be and index fields according to that. Also, remember than indexing fields like booleans is usually wasteful.

EshaanAgg:
@xoldyckk 
Read all your comments, reviews and the resources you can pointed to for pagination spec as well as indexing. I have updated all the models and typedefs as indicated by you. Please review!

palisadoes:
> @palisadoes Because according to current implementation, in the database same organization tag will be used for many different objects, it's not limited to only users. So, if it is needed the same tag can be used to fetch all objects related to that tag within a single screen. So, an admin can use non-sensical tags which only relate to users for events too, a tag is not related to only one type of object. It'll be the job of admins to remember this for sanely using tags for what they create it for.
> 
> Just wanted to let you know about this.

Whatever the current implementation in the DB, in future tags need to be categorized by the type of resource they are tracking so that we don't have this type of problem.

EshaanAgg:
@xoldyckk 

What do you think should be the database implementation now? Should be keep separate tables for each type of tag, like
```
organizationUserTag - List tags for classifying users
UserTag - Relational schema used to assign tags to users

organizationEventTag - List tags for classifying users
EventTag - Relational schema used to assign tags to events
```

This would require separate mutations though for each type of tags, as they would be now different types altogether.

xoldyckk:
@EshaanAgg Yeah, I guess. It's no big deal though, just needs repetition of the same logic we have for users, for events too. This is the most clean approach too, we won't have to do dirty stuff like seperating users and events or other objects.

EshaanAgg:
@xoldyckk 
Got it. I would read about the same in great detail. 

Currently I have updated the typeDefs and mongoose schema's to reflect the changes for only user-tags, and implement a generic Error Type which is not being consumed anywhere as off now. What do you think?

src/models/EventTag.ts
```diff
@@ -0,0 +1,32 @@
+import { Schema, model, PopulatedDoc, Types, Document, models } from "mongoose";
```
xoldyckk, line 1: Remove this file, add this in another PR

src/models/index.ts
```diff
@@ -18,5 +18,8 @@ export * from "./Organization";
 export * from "./Plugin";
 export * from "./PluginField";
 export * from "./Post";
+export * from "./Tag";
```
xoldyckk, line 21: Fix the imports.

src/typeDefs/inputs.ts
```diff
@@ -17,6 +17,12 @@ export const inputs = gql`
     title: String!
   }
 
+  input CreateUserTagInput {
+    name: String!
+    parentTagId: ID
```
xoldyckk, line 22: Alphabetical order.

src/typeDefs/types.ts
```diff
@@ -294,6 +303,50 @@ export const types = gql`
...
+    edges: [UserEdge]
+    pageInfo: PageInfo!
+  }
+
+  type UserEdge {
+    node: User
+    cursor: String!
+  }
+
+  type Error {
```
xoldyckk, line 345: Remove this for now. We'll think about errors schema at a later time.

xoldyckk (commented):
Try to keep fields inside `type`, `input`, `union` and `interface` alphabetical. Makes it easier to read.

xoldyckk:
@EshaanAgg This PR is too big. Revert the changes for resolvers and tests. Just make it typeDefs and mongoose models for now. Make new PRs for resolvers and tests, it'll be easier to review.

EshaanAgg:
Done @xoldyckk! Please review.

beingnoble03:
@xoldyckk @EshaanAgg a doubt here.
We have written
```ts
  userId: {
    type: Schema.Types.ObjectId,
    required: true,
    ref: "User",
  },
  tagId: {
    type: Schema.Types.ObjectId,
    ref: "OrganizationTagUser",
    required: true,
  },
```
in the `TagUserSchema `, which seems right as this Schema contains `userId` not an embedded document/
And, in the `Interface_TagUser`, we have written
```ts
  userId: PopulatedDoc<Interface_User & Document>;
  tagId: PopulatedDoc<Interface_OrganizationTagUser & Document>;
```
Here, it kinda feels wrong to use `userId` because, it now contains a Populated Document.
So, Is there something wrong with our naming? This doubt might be silly, but please clarify this once.

src/typeDefs/types.ts
```diff
@@ -347,4 +362,43 @@ export const types = gql`
...
+
+  type UserEdge {
+    node: User
+    cursor: String!
+  }
+
+  type UserTag {
+    _id: ID!
+    name: String!
+    organization: Organization!
```
xoldyckk, line 374: Make this nullable, all fields which make new network requests to get the data should be nullable, you don't want failed network request for a nested field to nullify the whole response object.

src/typeDefs/types.ts
```diff
@@ -229,6 +236,8 @@ export const types = gql`
     When paginating backwards, are there more items?
     """
     hasPreviousPage: Boolean!
```
xoldyckk, line 238: Add `hasNextPage` here, and please explain what `totalPages` `nextPageNo` and `prevPageNo` mean here? Connections don't work like page based pagination or am I missing something here?

EshaanAgg:
Right but that will only be true when we call the .populate method in out queries. As @xoldyckk stated earlier, we should always name fields as what they are storing, and not as what they will become after transformations such as population. 

PS. I completely agree it's weird. Even while writing resolvers, I have to do a step of the sort .map((tagUser) => tagUser.userId) to create a Users array! (and not userId array). It's weird, but there is no win-win situation here.

xoldyckk:
@beingnoble03 It's mongoose related quirk, and probably v5 specific. New mongoose versions probably have better approach to write model interfaces.

beingnoble03:
@xoldyckk it's pretty weird.
Just think of writing 
userId: User! in types.ts

src/typeDefs/types.ts
EshaanAgg: @xoldyckk 
Actually I had created a combined `PageInfo` type, which would contain attributes for both cursor based navigation (as we are implementing) and Offset based pagination which already exists in the codebase.

Though in hindsight, this is a practice, and I have defined a new type `CursorPageInfo` which only contains the attributes specific to the same.

src/typeDefs/types.ts
```diff
@@ -114,7 +122,7 @@ export const types = gql`
     title: String
     description: String
     createdAt: DateTime
-    organization: Organization!
+    organization: Organization
```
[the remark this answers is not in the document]
EshaanAgg: Haven't made any change to the existing schema, but made the Organization field on TagUser nullable. Probably the combined view on GitHub messing with you.

xoldyckk:
@EshaanAgg You changed it on `type Group`.

EshaanAgg:
My bad @xoldyckk! Updated the same.

xoldyckk (approved):

palisadoes (approved):
Hooray! After 135 comments we can merge.

xoldyckk:
@palisadoes The resolvers and tests are left. The initial changes @EshaanAgg made regarding resolvers and tests at the start of PR would no longer work because the mongoose models and graphql typeDefs had many changes, so I've told him to open new PRs for them.

xoldyckk:
@palisadoes Nothing is added that will break the app. The resolver that's added is for validating positive scalar int value. It's to make sure that clients enter positive numbers only for fields which have `PositiveInt` as their type in graphql schema.

palisadoes:
That's what I thought, thanks

PR #1168, opened 2023-03-12
image upload resolver init

### LINKED ISSUE #1138: Implementing image resolvers where needed.
**Describe the bug**
Now that the Encoded Image upload is available in the server the resolver to resolve path of the image can be implemented.

Keep in mind this is a sister issue and go hand -in -hand:
https://github.com/PalisadoesFoundation/talawa-admin/issues/538

**To Reproduce**

1.
2.
3.
4.

**Expected behavior**

**Actual behavior**

**Screenshots**

**Additional details**

kb-0311:

**Issue Number:**

Fixes #1138 

**Snapshots/Videos:**

**Summary**

**Other information**


kb-0311:
This is still a PR in progress 

Also its sister PR for admin portal is this one-

https://github.com/PalisadoesFoundation/talawa-admin/pull/677

kb-0311:
Screenshot from 2023-03-12 10-57-28
After this PR the admin portal will be able to use image upload functionalities perfectly .

Screenshot from 2023-03-12 11-00-26
Screenshot from 2023-02-24 12-54-38
Screenshot from 2023-03-12 11-00-26
Screenshot from 2023-03-12 10-57-13
Screenshot from 2023-03-12 11-19-35


The manual testing was a success just need to write the unit tests

xoldyckk:
@kb-0311 

Just so you know, this isn't a flexible approach for sending back static files to the clients and I'm not talking about the base64 thing here. Right now you're sending back just urls(I assume) to images, but in actuality this isn't how static files should be shared over graphql.

Though it depends on the file type and how much granular control you want to provide to the clients, different file types have different graphql types defined in schema, and they're used to send back client complete data about the file.

For example, here's a basic example of how a flexible image can be sent to client:-

```ts
type Image {
  id: ID!
  url: String!
  alt: String
  title: String
  metadata: ImageMetadata!
}

type ImageMetadata {
  width: Int!
  height: Int!
  format: ImageFormat!
  fileSize: Int!
  mimeType: String!
}

enum ImageFormat {
  JPEG
  PNG
  GIF
  WEBP
}
```
Though this is still not as highly complex implementation as big companies like google, netflix etc . use. 

The clients can then transform the image url with fields from extra metadata about the image, to flexibly query for images according to their needs. The approach is similar for other file types you want to send like videos, documents, sounds etc. The possibilities are endless, just wanted to let you know.

xoldyckk:
@kb-0311 Also, since you're fixing up updateOrganization would you like to fix these things?

Accept a single argument named `input` of type `UpdateOrganizationInput`, include the organization's id within that and make the id required. Also make the input required in mutation definition like this `input: UpdateOrganizationInput!`. It doesn't make sense to make a mutation request without the input object. Also, notice how things are alphabetically ordered in the input? It's a great pratice to organize stuff alphabetically except for unique fields and id fields, they should be put on top, after that everything should be alphabetically ordered.
```
  input UpdateOrganizationInput {
    _id: ID!
    description: String
    isPublic: Boolean
    location: String
    name: String
    visibleInSearch: Boolean
  }
```
I'm asking this because mutation receiving a single argument named input is the convention in graphql. It's a good practice to stick to conventions. Also, currently eslint-graphql is not correctly configured to enforce these constraints, but in the future it'll be easier to do so, if there are less things to fix. Take a look at the rules, you'll see the benefits it provides:- https://the-guild.dev/graphql/eslint/rules. There's also pre-defined configs and they enforce some of the common rules.

kb-0311:
@xoldyckk Yep , I completely agree with you regarding the image input type and giving the clients more control over what data they need to send to the API. Your approach is a lot better and type-safe. But actually my exams are starting from tomorrow and will end on 18th. So trying not to make a lot of changes that would require extra care and too much time in implementing the changes.

But I think this approach should be implemented for image and video files. I will open an issue after 18th maybe to complete that along a sister PR with the changes on the client side.

Regarding the Update User type yes, I will try to do that in this PR.

src/typeDefs/mutations.ts
```diff
@@ -157,8 +157,11 @@ export const mutations = gql`
 
     updateLanguage(languageCode: String!): User! @auth
 
-    updateOrganization(id: ID!, data: UpdateOrganizationInput): Organization!
-      @auth
+    updateOrganization(
+      id: ID!
+      data: UpdateOrganizationInput
+      file: String
```
xoldyckk, line 163: Instead of file, naming it image, avatar etc., would be better. For now its type is `String`, but later on it will be changed to `File` I suppose. 

Also, include this one inside the input field too, don't need a seperate argument for it.

xoldyckk:
@kb-0311 Will you be making the necessary changes for `createPost`, `updateOrganization` and `updateUser` within this PR?

src/typeDefs/mutations.ts
kb-0311: @xoldyckk I thought the file type is there to denote both image and video files that is. 

But changing its name OR keeping it inside the input type would need changing in a lot of places within multiple resolvers where if(args.file ) check is there and changing the queries and mutations in the admin-portal too in the client, it will require a lot of work. And yk the exam thingy and all.

kb-0311:
@xoldyckk what necessary changes are you taking about in create Post and the other mutation if I change the updateOrganizationInput? Wouldn't the codegen handle that?

src/typeDefs/mutations.ts
kb-0311: But you are right though the file input could be handle within the `data:` field itself so in the after 18th PR I will do that too.

xoldyckk: Hmm, yes I guess it can be a lot of work if you'll also need to change the tests, and also make the changes to clients and thier tests as well.

xoldyckk:
> @xoldyckk what necessary changes are you taking about in create Post and the other mutation if I change the updateOrganizationInput? Wouldn't the codegen handle that?
> 
> Also closed by mistake.

Basically what you said above in the review conversation. I guess you can leave it for now. I'll review the PR as it is.

jadeharr:
@kb-0311 Please make sure that the code coverage does not decrease.

kb-0311:
@xoldyckk hmm this is weird , in my system the tests are passing.\

What could be the source of anamoly?
Screenshot from 2023-03-13 18-12-13

kb-0311:
Nevermind , got the issue , completed the tests , Now it should be okay.

kb-0311:
> Basically what you said above in the review conversation. I guess you can leave it for now. I'll review the PR as it is.
> 
> But since you're adding the new image argument to `updateOrganization`, at least fix that one.

Sure, right now the id field for updateOrganization exists in the mutation arguments though but outside the data field, You want to move it from the mutation argument to the data field? If yes the mutation types will also need to be changed, right?

xoldyckk:
@kb-0311 Yes, also `data` should be named `input`.

palisadoes:
We really need this completed.

xoldyckk:
@kb-0311 You going to make the changes for `updateOrganization` within this PR or later on? Tell me right now and I'll review this PR.

Btw those issues @palisadoes linked just now, what about video upload, is it gonna work perfectly with base64 encoding approach?

kb-0311:
@xoldyckk You can carry on with the review I will submit another PR later on to completely improve the types everywhere.

The base64 Encoding will not work well for videos at all.

palisadoes:
1. Yes, let's get this merged. It's blocking others.
1. If certain media types aren't supported for now, we can fix that later as part of a GSoC idea.
1. Consider sending a rejection notification for unsupported media types if applicable.

kb-0311:
@palisadoes 
Yes sure.

The rejection of unsupported media types is done by only accepting file with png or jpg extensions both in the front end and backend

DMills27 (approved):

palisadoes:
@kb-0311 @xoldyckk Dominic approved this. Can we merge?

src/resolvers/Mutation/createPost.ts
```diff
@@ -108,5 +109,10 @@ export const createPost: MutationResolvers["createPost"] = async (
   }
 
   // Returns createdPost.
-  return createdPost.toObject();
+  return {
+    ...createdPost.toObject(),
+    imageUrl: createdPost.imageUrl
+      ? `${BASE_URL}${uploadImageFileName}`
+      : undefined,
```
xoldyckk, line 116: Make the value after `:` null. `undefined` is kinda a wack thing in javascript which should be avoided whenever we can. For nullable values `null` should be explicitly used. On client side, `imageUrl` can either be a `string` or `null` if the client asked for it in their graphql query explicitly. Graphql would never return `undefined`.

src/resolvers/Mutation/updateOrganization.ts
```diff
@@ -22,13 +23,22 @@ export const updateOrganization: MutationResolvers["updateOrganization"] =
...
+    }
+
     return await Organization.findOneAndUpdate(
       {
         _id: organization._id,
       },
       {
         $set: {
           ...args.data,
+          image: uploadImageFileName || undefined,
```
xoldyckk, line 41: I've heard that prior to mongoose v6, if you use `undefined` in update fields, mongoose will set those fields to null in database instead of just ignoring them. If this is an intended behaviour ignore this comment.

src/resolvers/Mutation/updateUserProfile.ts
```diff
@@ -50,46 +51,29 @@ export const updateUserProfile: MutationResolvers["updateUserProfile"] = async (
...
-    ).lean();
-  }
+    },
+    {
+      new: true,
+    }
+  ).lean();
+  updatedUser!.image = updatedUser?.image
+    ? `${BASE_URL}${updatedUser?.image}`
+    : undefined;
```
xoldyckk, line 76: Same as comment 1.

src/resolvers/Query/organizationsMemberConnection.ts
```diff
@@ -58,13 +59,15 @@ export const organizationsMemberConnection: QueryResolvers["organizationsMemberC
       users = usersModel.docs.map((user) => {
         return {
           ...user,
+          image: user.image ? `${BASE_URL}${user.image}` : undefined,
```
xoldyckk, line 62: Same as comment 1.

src/resolvers/Query/organizationsMemberConnection.ts
```diff
@@ -58,13 +59,15 @@ export const organizationsMemberConnection: QueryResolvers["organizationsMemberC
...
           password: null,
         };
       });
     } else {
       users = usersModel.docs.map((user) => {
         return {
           ...user._doc,
+          image: user.image ? `${BASE_URL}${user.image}` : undefined,
```
xoldyckk, line 70: Same as comment 1.

src/resolvers/Query/postsByOrganization.ts
```diff
@@ -20,4 +21,11 @@ export const postsByOrganization: QueryResolvers["postsByOrganization"] =
       })
       .populate("creator", "-password")
       .lean();
+
+    const postsWithImageURLResolved = postsInOrg.map((post) => ({
+      ...post,
+      imageUrl: post.imageUrl ? `${BASE_URL}${post.imageUrl}` : undefined,
```
xoldyckk, line 27: Same as comment 1.

src/resolvers/Query/postsByOrganizationConnection.ts
```diff
@@ -46,6 +47,7 @@ export const postsByOrganizationConnection: QueryResolvers["postsByOrganizationC
     const posts = postsmodel.docs.map((post) => {
       post.likeCount = post.likedBy.length || 0;
       post.commentCount = post.comments.length || 0;
+      post.imageUrl = post.imageUrl ? `${BASE_URL}${post.imageUrl}` : undefined;
```
xoldyckk, line 50: Same as comment 1.

src/resolvers/Query/user.ts
```diff
@@ -25,6 +25,7 @@ export const user: QueryResolvers["user"] = async (_parent, args, context) => {
   // This Query field doesn't allow client to see organizations they are blocked by
   return {
     ...user!,
+    image: user!.image ? `${BASE_URL}${user!.image}` : undefined,
```
xoldyckk, line 28: Same as comment 1.

src/resolvers/Query/users.ts
```diff
@@ -33,6 +33,7 @@ export const users: QueryResolvers["users"] = async (_parent, args) => {
     return users.map((user) => {
       return {
         ...user,
+        image: user.image ? `${BASE_URL}${user.image}` : undefined,
```
xoldyckk, line 36: Same as comment 1.

xoldyckk (commented):
@kb-0311 @anshgoyalevil Made a few changes to context few days ago. Check them out here in the context definition for apollo server:- https://github.com/PalisadoesFoundation/talawa-api/blob/develop/src/server.ts

The BASE_URL you're using is not dynamic. On the context object, now there a `apiRootUrl` variable which can be used, I'd suggest looking at it and using it instead of `BASE_URL`.

kb-0311:
@xoldyckk The `:null` assignment instead of undefined fives this error
Screenshot from 2023-03-14 10-44-17
I guess it is because of the post model interface thing.

Do you want me tochange these too?

Screenshot from 2023-03-14 10-48-02


Yes gonna look at apiRootUrl thing .

kb-0311:
the context is undefined when accessed in the resolver. I think it is because it requires an express request to initialize it first. It could be done, but then re writing the tests again to make use of the express request types and the mock configure it :skull:. Tbh really do not have time to do that in the exam week 

How about we let the newer outreachy people do those changes? It will also give those guys a relatively okay issue to start off with according to what @palisadoes was talking about. If no one takes up I will fix those changes with the types PR fix on the 18th / 19th I was talking about previously for the UpdateOrganizationInput.

Screenshot from 2023-03-14 11-58-47
Screenshot from 2023-03-14 11-58-38

xoldyckk:
@kb-0311 `undefined` is used for values which are uncertain, for example result of a network call may arrive at a later point in time. So, its type will be `NetworkCallResult|undefined`. Database fields for example have a concept of `null` but not `undefined`.

Since, mongodb has no enforced schema at the database level, `undefined` is possible. Here are the possible scenarios:-
1. When documents in the same collection may or may not have a field, but if the field exists it must not be null. In this case the field can be undefined.
2. When documents in the same collection may or may not have a field, but if the field exists it can be null. In this case the field can be both undefined and null.
3. When documents in the same collection must have a field, but it cannot be null. In this case the field cannot be null.
4. When documents in the same collection must have a field, but it can be null. In this case the field can be null.

`undefined` in javascript can be implicit or explicit. For example:-
```ts
// explicit

const obj = { myField: undefined };
// This will be undefined
obj.myField


// implicit

const obj = { };
// This will be undefined, even though the field is not explicitly present and defined in obj
obj.myField
```

So, how you define those interfaces for mongoose models depends on the factors mentioned above. But interfaces by themselves have nothing to do with how you update you existing documents. In the input object client passes to the server, they may want to update fields partially, so they might not pass some arguments. In graphql resolver inside the input object, those fields will be `undefined`. If you set `undefined` as value for a field in mongoose version prior to v6, it'll just make the fields null. Client didn't want the field to be null, but you unintentionally did that.

noman2002 (commented):
Fix the conflicting files.

anshgoyalevil:
@kb-0311 

You can add the apiRoolUrl to the context while writing the tests.

For example:

```ts
const context = {
    userId: testUser!._id,
    apiRootUrl: "http://localhost:4000",
};
```

You don't need to mock express requests.

palisadoes:
@kb-0311 Please fix the conflicting files.

kb-0311:
@xoldyckk @palisadoes sorry for the late reply, My mid sem exams are going on currently so was a little inactive as I had mentioned here.https://github.com/PalisadoesFoundation/talawa-admin/pull/677#issuecomment-1466877772

@xoldyckk Ok got it gonna change it to null and use the BASE_URL for testing purposes for apiRootUrl thing

So I will also need to change the interface `Post` or other interfaces relevant here right? Basically change the `image : string|undefined |null` or `string|null`?

xoldyckk:
@kb-0311 If the client wants to make a field null in database they have to explicitly provide the value null in their graphql query. If nothing is provided(it is undefined), the mutation should not affect that field(it should not assume that undefined = set the value in database to null).

If the interfaces require fix, make the fixes.

kb-0311:
@xoldyckk changed the undefined to null.

I made use of apiRootUrl in mutations but Queries do not get context passed onto them currently so will need to do that too, in the next PR

Regarding our previous conversation about types for Image and ImageMetadata and UpdateOrganizationInput you can add a comment here on what could be changed and I will do that as a whole in that PR.  


You can safely approve and merge this.

xoldyckk:
@kb-0311 What do you mean queries don't get context passed to them? It's an overloaded function, you can choose whether you want to include `parent, args, context, info` if the resolvers needs them. Those query resolvers had no use of context object that's why it was not included in function definition. You can simply include it like this 
```ts
(parent, args, context, info) => {
  ...use context object here
}
```

Regarding the talk about representation of static file types in graphql schema, that depends on how you want to serve those files to clients. For example if you want the clients to only use images with same type, size, resolution, quality, format etc., then there's no problem with a simple `imageUrl`. But that's not how things work. You might need a very small sized image for showing a small profile picture of a user when you're displaying their profile in a small card, but a very large one when you're displaying their profile in full page.

Whether you push the work for later or now is for you think about. If the client side developers ask that extra information about files(images for example) is needed on client side to have more flexible querying options for files, you'll have to refactor everything again.

palisadoes:
@kb-0311 @xoldyckk Guys, wow long before we get a resolution to this PR? It's been submitted for a week and it has been blocking development in other areas. For example, here is the latest discovery.

- https://github.com/PalisadoesFoundation/talawa-admin/issues/756

The aim outlined a few weeks ago was to have a sub-optimal stopgap measure that would work for most cases and then a GSoC project would find the long term solution. We are at 45 comments and counting.

kb-0311:
@xoldyckk used the context in the queries too. 

For now, I am leaving the static site serving as is the image uploads will be refactored entirely anyways no point in doing all that right now.

xoldyckk (approved):
EHHH! This is so much hacked together way of making this functional. Well whatever.

@palisadoes Currently only image upload will work. And that too for only 1 single image. Multiple images cannot be uploaded. 

For something like post multiple images can't be uploaded and videos are out of the question.

If you want the perfect thing many things will need to be reimplemented from ground up and they will break both talawa-admin and talawa-mobile.

palisadoes:
This will need to be resolved with GSoC. Thanks.

palisadoes (approved):

palisadoes:
@xoldyckk

1. We have this GSoC idea that includes fixing image uploads. https://docs.talawa.io/docs/internships/internship-ideas/#general-basic-functionality 
2. Does this image upload work justify a separate 350 hour project? 
    1. Can it be handled within the current project and meet the other goals?

kb-0311:
@xoldyckk Yes I completely agree with you, Check out the file-upload channel on slack although you would need to scroll a lot I had said.

I know this is not the best method of doing it (with respect to Graphql plus encoded image) what @palisadoes  had asked me was to implement an encoded image thingy which would allow the client apps to at least make use of image uploads which was completely dysfunctional and that too as quick as possible and not worry about kind of optimized approach for now.

palisadoes:
@kb-0311 What are your thoughts on my previous comments?

xoldyckk:
@palisadoes It's not about fixing image uploads. It's about fixing any kind of upload. Different files have different formats and metadata associated to them. With proper upload functionality you can transfer just about anything over the network in the way it's supposed to be. What you accept as upload on the server depends on your needs. Base64 is not meant for file uploads.

palisadoes:
@kb-0311 Assuming we were to support an infinite range of file types, what would your answer be to my previous question? We need to start aligning GSoC to a working system.

kb-0311:
@palisadoes @xoldyckk yes, I had previously also said that a better approach for file uploads would be sending the actual file as a graphql-upload scalar from the clients and resolving them as an upload-scalar in the server before anything so what @xoldyckk is saying here would be implemented. Because then we will be sending over a general file supporting all general types of files on the client side
>It's about fixing any kind of upload. Different files have different formats and metadata associated to them. With proper upload functionality you can transfer just about anything over the network in the way it's supposed to be.


 But for that we will have to migrate to graphql-yoga first which has been delayed for various reasons but I think it is high time that it should be given a higher priority now.

We can even send over the file metadata by creating a type for a File where in the metadata and upload scalar will be defined. https://the-guild.dev/graphql/yoga-server/docs/features/file-uploads

Now the burden of resolving the file types and storing them in the API server file system would be on the API part. That way the burden to encode would not be on the client's side.

@palisadoes regarding your GSoC question. I will be contributing to that idea you linked this Year. A long-term file uploads resolving idea does deserve its own project it may look easy but it really is not and certainly could not be practically covered within existing ideas this year. Because as you had said previously we will also need a script that when run will make sure all the future uploads from that point will be on the cloud.

kb-0311:
@palisadoes But I, personally, would be okay with at least getting started with resolving the File scalar in graphql after yoga migrations during that Project timeline or even before that if the migration is done fast. But this will only be an optional change. 

That way the long-term project to make a scalable and consistent File upload utilities can be easier to implement next year

kb-0311:
@palisadoes No, but it does not have to be.It can be done before GSoC if a few more contributors start off again. That channel became cold because everyone there had started working on other issues. I think now it could be the time to revive it again.

@anshgoyalevil @EshaanAgg @beingnoble03

palisadoes:
It would be excellent to restart that. It started off so strongly.

palisadoes:
I've added this to my list of GSoC 2024 ideas

EshaanAgg:
I had created a branch on my fork for the yoga migration and am definitely willing to work on it!
A bit of support from other contributors and we can do it quickly and effectively.

kb-0311:
@palisadoes Cool that's great. 

Yeah will try to get the boys back on the migration issue.

kb-0311:
@EshaanAgg sure lets finally have the meet for that

PR #1179, opened 2023-03-15
Add field level resolvers for UserTag Models [Fixes part of #1126]

EshaanAgg:
Feature

**Issue Number:**

Fixes part of #1126 

Yes

**Snapshots/Videos:**
NA

NA

**Summary**
In the previous PR [#1158], types were introduced for the `UserTag` type. Due to the support for the same, we need the followinf field level resolvers, out of which some are implemented in this PR for easier review:
- [X] `User/tagsAssignedWith`
- [x] `Organization/userTags`
- [X] `UserTag/organization`
- [X] `UserTag/childTags`
- [X] `UserTag/parentTag`
- [X] `UserTag/usersAssignedTo`

No

**Other information**
- This PR introduces the first examples of cursor based navigation to the repository, thus I would like a general review before I can complete the implementation of other field level resolvers as well.
- Due to the highly nested nature of types in the implementation of this feature, in some resolver tests the typescript inference is wrong/restrictive. Thus in such cases the use of `// @ts-ignore` has been made. 

#### Important
- Changed the `typeDefs` to add a required argument called `organizationId` for `User/tagsAssignedWith` field resolver, as the tags of an organization are private to an organization, and thus it makes sense to only allow queries to get the tags for a user for one specific organization at a time, and not all tags across all organizations at once. 

Yes


src/resolvers/UserTag/organization.ts
```diff
@@ -0,0 +1,16 @@
+import { UserTagResolvers } from "../../types/generatedGraphQLTypes";
+import { OrganizationTagUser } from "../../models";
+
+export const organization: UserTagResolvers["organization"] = async (
+  parent
+) => {
+  const organizationObject = await OrganizationTagUser.findOne({
```
xoldyckk, line 7: This implementation is incorrect. You have the `organizationId` field available on `parent`. Use it to directly fetch the organization object from database.

xoldyckk (commented):
@EshaanAgg You've used parent._id everywhere in nested relations to fetch the related object(s). That's not how it's done. Read the review for `organization` field I left below.

src/resolvers/UserTag/organization.ts
EshaanAgg: image
Using the same gives the following typescript error, which I think so is correct as parent here is `UserTag type` and not the `OrganizationTagUser model`, which does not have any `organizationId` field, but rather only an `organization` field which needs to be resolved. Or am I mistaken?

xoldyckk: @EshaanAgg You haven't mapped `UserTag` in graphql schema to `OrganizationUserTag` interface. Do that and typescript won't complain.

EshaanAgg: @xoldyckk Done! Worked like a charm.

EshaanAgg:
@xoldyckk A general note on the cursor based pagination implemented:

While paginating forwards:
1. In the first request, the user is expected to only send the `first` argument (say `10` in this example). This would return a list of all the first 10 objects queried.
2. Now if the returned object has `hasNextPage` as false, then the pagination ends. The user can now paginate backwards if it has `hasPreviousPage` to be true, by sending a request with `before` set to `startCursor` of the current page.
3. If the returned object has `hasNextPage` as true, then the user can paginate forwards by sending a request with `after` set to the `endCursor` of the page.

Similar logic works for pagination backwards, by just making obvious switches!

xoldyckk:
@EshaanAgg Do one thing, make seperate PRs for each of the resolvers, it'll be easier to review and also point out your mistakes. Make the PRs for easier resolvers first, later on for the connections.

EshaanAgg:
@xoldyckk Done

EshaanAgg:
@xoldyckk 
Why did you close the PR? It still has code and tests for two easy field level resolvers.

xoldyckk:
@EshaanAgg just make a seperate PR for them.

PR #1187, opened 2023-03-16
Add field level resolvers for `UserTag/usersAssignedTo` [Fixes part of #1126] 

EshaanAgg:
Feature

**Issue Number:**

Fixes part of #1126 

Yes

**Snapshots/Videos:**
NA

NA

**Summary**
NA

No

**Other information**
- This PR introduces the first examples of cursor based navigation to the repository, thus I would like a general review before I can complete the implementation of other field level resolvers as well.
- Due to the highly nested nature of types in the implementation of this feature, in some resolver tests the typescript inference is wrong/restrictive. Thus in such cases the use of `// @ts-ignore` has been made. 

Yes


EshaanAgg:
@xoldyckk Can you please review this PR first, as it the first PR that introduces cursor paged pagination. Once this is approved, adding other field resolvers would be trivial, as they would largely follow the same design patterns and coding style.

EshaanAgg:
It's been 4 days since the PR was opened. I know that the maintainers must be super busy right now, but just wanted to keep the conversation active here.
Is there anything that I can do that would perhaps help in the review process?
@xoldyckk @palisadoes

noman2002 (dismissed):

xoldyckk:
Well anyway I'm going to outline the biggest flaw with your approach. You are fetching every single user that is assigned to a tag in a single db request. This completely goes against the concept of pagination. Pagination isn't only for providing data in little chunks to client, it's also to get data in little chunks from the database.

Now flaws regarding resolver arguments:-

1. Either argument `first` or `last` must be provided. If they're not the query should fail.
2. If both `first` and `last` are provided, the query should fail, there's no use case currently where both the arguments should be accepted in a single query.
3. Arguments work in pair here. `first/after` are a pair and `last/before` are a pair. Argument `after` is optional, but if it is provided `first` must also be provided. Clients cannot provide `before` with `first`. Argument `before` is optional, but if it is provided `last` must also be provided. Clients cannot provide `after` with `last`.
4. `after` and `before` are cursors. Cursors are base64 encoded strings that contain information to uniquely identify an edge in a connection. Edge here would be the user edge(in graphql schema). Primary id(mongodb _id) can be used as a cursor, though usually it's a combination of primary id and createdAt field.
5. For getting information about `hasNextPage` and `hasPreviousPage`(depending upon whether you're forward or backward paginating), you can increase the number of items you're fetching from database by 1. If that extra item is fetched too, it means you have a next page of data(or previous page).

**For sometime in the future when a good validation system is decided for resolver arguments:-** You're not checking arguments `first` and `last` for minimum/maximum. The client cannot just query with `0` as the value for either `first` or `last`. Accordingly, client cannot query with a big integer like `999999` as the value for either `first` or `last`. Idk if the custom graphql scalar `PositiveInt` includes `0` as a positive integer, but if it does the query should fail. `first` and `last` dictate how many items are returned in a single query, their values should be constrained. The client cannot just ask to get 0 items or 99999 items in a single request.

EshaanAgg:
Yes I'm totally aware of how detailed your reviews are, and am very appreciative of the time and energy you spend in doing so.

The point of the message wasn't to pester you, but again just keeping the conversation active.

EshaanAgg:
@xoldyckk Updated the pagination code to implement the validations you requested. What do you think?

PS. Haven't added the tests for the changes yet. Will do so after making sure we are on the same page about the validation workflow.

EshaanAgg:
@palisadoes @xoldyckk @noman2002 
I have created a new validation file (library) for all the checks that were recommeded by @xoldyckk in the discussion, and have added the tests for the same as well.

xoldyckk:
@EshaanAgg Once [this PR](https://github.com/PalisadoesFoundation/talawa-api/pull/1213) is merged make use of the factory object.

EshaanAgg:
@xoldyckk Can you elaborate a bit more as to how I can use the factory object in my implementation?
Currently the factory function just seems to return a blank connection object (which has no edges).

xoldyckk:
@EshaanAgg read the description I've written for that function.

EshaanAgg:
Hmm got it!
But why are we first creating a blank object, and then populating it with data fetched, whereas in the current implementation we are building the object to be returned directly just before it is returned?

xoldyckk:
@EshaanAgg Because the connection object can be returned at many steps. If you get an empty list from the database you can return the connection object immedietely, instead of doing all the redundant logic you're currently doing and after that you're returning the final thing.

EshaanAgg:
Got it @xoldyckk!

xoldyckk (commented):
@EshaanAgg You have to validate whether the cursor that client has sent is valid or not. By that I mean does a database objects exists which references that cursor. If not you return a suitable error. The simplest way to do this would be to include the object related to that cursor in the db query. So, you include 2 more objects than the number specified by clients. First object would be to make sure object with provided cursor exists, and last object would be to determine whether next page exists.

Another thing is right now you're sorting by `userId` which is weird. Normally you should sort by `createdAt` field for the object. We don't have an explicit `createdAt` field in `TagUser`, but mongodb object ids have timestamp encoded in them. So, if you sort a mongodb object by it's primary object id, it'll sort the collection the same way something like a `createdAt` field would.

EshaanAgg:
@xoldyckk 
Got the point about the cursor validation and would update the code to reflect the same. 
As for the sorting by `userId`, the logic that I'm using for the same is that since we are returning user objects, it makes more sense to return the objects sorted by the `createdAt`/`_id` of these users. If we use the `_id` of the UserTag model, we would be sorting on the basis of the order in which the users were assigned the tag, latest one last.
To be personally the former one made a bit more sense, but would change the implementation if you feel the later is infact a better design.

EshaanAgg:
Got it!

EshaanAgg:
@xoldyckk Made the required changes. Can you please review now?

src/resolvers/UserTag/usersAssignedTo.ts
```diff
@@ -0,0 +1,131 @@
...
+    allusersAssignedTo = allusersAssignedTo!.reverse();
+  }
+
+  // Create edges from the fetched objects
+  connectionObject.edges = allusersAssignedTo!.map(
+    (tagUser) =>
+      ({
+        node: tagUser.userId,
+        cursor: tagUser._id.toString(),
+      } as Interface_ConnectionEdge<Interface_User>)
```
xoldyckk, line 122: Is this `as` assertion necessary? Is typescript complaining without it?

One problem could be that `userId` type here is `string` instead of `Interface_User`? You can assert the same thing on `tagUser.userId`, then you won't need to import `Interface_ConnectionEdge`.

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -1,6 +1,6 @@
 import { ConnectionPageInfo } from "../types/generatedGraphQLTypes";
 
-interface Interface_ConnectionEdge<T> {
+export interface Interface_ConnectionEdge<T> {
```
xoldyckk, line 3: If this export isn't needed, remove it, refer to comment above.

xoldyckk (commented):
It looks kind of good enough, some shortcomings will probably be caught while testing against different input.

One thing that's problematic is the code duplication and it looks really unclean.

src/resolvers/UserTag/usersAssignedTo.ts
EshaanAgg: @xoldyckk 
image
This is typescript error that I was facing without the as assertion. I can't seem to resolve the same.

src/utilities/graphqlConnectionFactory.ts
EshaanAgg: Sure will do!

EshaanAgg:
> It looks kind of good enough, some shortcomings will probably be caught while testing against different input.
> 
> One thing that's problematic is the code duplication, and it looks really unclean.

I tried brainstorming to find a way to remove the code duplication for thee forward and backward pagination, but I couldn't do so without making the code, even more, messier, with soo many in-line conditionals that code readability goes out of the picture. 

Hopefully, future contributors would be able to resolve this and make this code better.

src/resolvers/UserTag/usersAssignedTo.ts
xoldyckk: @EshaanAgg Your first problem is using `!` to assert that something cannot be null, it should be sparingly used and instead there should be strict type checking which exists at runtime.

And in your function your map function you're not returning anything.

This is wrong:-
```ts
=> 
{
  node: tagUser.userId,
  cursor: tagUser._id
}
```
Do it like this:-
```ts
=> 
( {
  node: tagUser.userId,
  cursor: tagUser._id
} )
```

EshaanAgg: Ahh you are right! Made a typing error here. Fixed the same and typescript stopped complaining. 
Thanks for the help @xoldyckk!
Have pushed the changes in the lastest commit.

jordan-barrett-jm:
Is this ready to be merged?

xoldyckk:
@EshaanAgg the implementation looks correct. Add more comments on why n+2 items are being fetched, and also check for grammatical errors/incorrect words in the comments, i noticed a few.

EshaanAgg:
Done @xoldyckk.

EshaanAgg:
@xoldyckk Can you please check the comments once whenever you are free?

xoldyckk:
@EshaanAgg There are still mistakes in comments, please go through them properly, write them like you're writing english essays, with care. You can start working on writing the tests for this resolver, I think we will get to see a few shortcomings of the current implementation. For tests you will need to create a list of users, tags, organizations and tagUsers, but don't make it too big something like 5 items for each, and 10-20 tagUsers would suffice. We just need to check how the result is sorted when using first/after and last/before.

EshaanAgg:
@xoldyckk Fixed the mistakes!
Also, I have already added the tests for this resolver and updated their descriptions in detail to convey precisely what the test is doing. The current implementation works by creating two tag objects, one tag (`testTag`) is then assigned to 5 different users, and the other tag object (`randomTag`) is assigned to none.

src/resolvers/UserTag/usersAssignedTo.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const connectionObject = graphqlConnectionFactory<Interface_User>();
+
+  // Forward pagination
+  if (args.first) {
+    // Fetch the users
+    allusersAssignedTo = await TagUser.find({
+      ...(args.after && {
+        _id: {
+          $gte: args.after,
```
xoldyckk, line 26: technically both `_id` and `userId` can be used for filtering here, but `userId` would make more sense here, as that's what is intended for.

src/resolvers/UserTag/usersAssignedTo.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    // Reverse the order of the fetched objects as according to Relay Specification, the order of
+    // returned objects must always be ascending on the basis of the cursor used
+    allusersAssignedTo = allusersAssignedTo!.reverse();
+  }
+
+  // Create edges from the fetched objects
+  connectionObject.edges = allusersAssignedTo!.map((tagUser) => ({
+    node: tagUser.userId,
+    cursor: tagUser._id.toString(),
```
xoldyckk, line 126: Same as above comment.

xoldyckk:
@EshaanAgg I think the more resolvers that are implemented like this, the more dirtier and boilerplatey implementing connections will be. There's been talk of implementing more more connection resolvers in talawa-api. Following this approach of implementing connections in other new resolvers too will result in sphagetti code and quickly grow out of hand. @kb-0311 has been working on thinking about a future proof and flexible approach for communicating errors to the clients.

The things you should think about are having utility functions that parse/validate/transform incoming connection arguments. Each connection has the four arguments `first`, `after`, `last` and `before` in common. So, the utility function logic for them will be the same in each resolver. The difference is in the sort/filter arguments these connections will accept. For each resolver sort/filter argument can be fundamentally different and so will be the logic for utility functions that parse/validate/transform them.

I'd like this PR to be the reference point for implementing additional connections, so I'd like the best approach to be followed here from the start. So, please take your time and think of possible solutions.

Share your thoughts [here](https://github.com/PalisadoesFoundation/talawa-api/discussions/1239).

src/resolvers/UserTag/usersAssignedTo.ts
EshaanAgg: Earlier the implementation worked on `userId` as the cursor itself, but in [the previous comments](https://github.com/PalisadoesFoundation/talawa-api/pull/1187#issuecomment-1481407955) you had mentioned that we should be sorting by the `createdAt` / `_id` property of the records. That's why I had shifted to using the record `_id` as the cursor.

src/resolvers/UserTag/usersAssignedTo.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    // Populate the page pointer variables
+    if (allusersAssignedTo!.length === args.last + 1) {
+      connectionObject.pageInfo.hasNextPage = true;
+      allusersAssignedTo!.pop();
+    }
+
+    // Reverse the order of the fetched objects as according to Relay Specification, the order of
+    // returned objects must always be ascending on the basis of the cursor used
+    allusersAssignedTo = allusersAssignedTo!.reverse();
```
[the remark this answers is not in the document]
EshaanAgg: The logic is incorrectly implemented. There is no need to reverse the list objects. Got it. Will update the same!

src/resolvers/UserTag/usersAssignedTo.ts
xoldyckk: @EshaanAgg wasn't that about sorting by `userId`? Here I'm talking about filtering by `userId`. Filtering using `tagId` and `userId` would work because there combination is unique. Filtering using `tagId` and `_id` would work too, because `_id` by itself is unique, so using `tagId` serves no purpose here anymore.

Technically you only have the convinience of using `_id` because mongodb auto generates it, it's not necessary to have `_id` field for this because this document will only ever be used in context of `tagId` and `userId`.

EshaanAgg: @xoldyckk 
But shouldn't we be filtering and sorting by the same property?
Otherwise won't be have a conflict as to which property should be the cursor?

EshaanAgg: If we are to implement the sorting by `_id` and filtering by `userId`, then in the request, we would have access to only one of these properties as the cursor. 
And obviously it can't be ensured that if two documents in the `UserTag` schema (say d1 and d2) have d1.id >= d2.id, then the same imply d1.userId >= d2.userId.

EshaanAgg:
Got it @xoldyckk. There are a lot of moving parts here, and specially the typescript support for the implementation makes the things way trickier. I'll read about the same and come up with a suitable approach.

EshaanAgg:
@xoldyckk @kb-0311 
I have updated this PR to use a general `GraphQLConnectionFactory` to generate connections for `UserTag/usersAssignedTo` and `UserTag/childTags` fields! I have also introduced the `GraphQL Errrors` as was suggested by @kb-0311. Can you please review the same and let me know your views!

PS. Will be adding tests later for all the changes.

EshaanAgg:
@xoldyckk Can you review this PR and give your views on the implemented general graphql connection factory?

xoldyckk:
> @xoldyckk Can you review this PR and give your views on the implemented general graphql connection factory?

I'll take a look.

EshaanAgg:
@xoldyckk What do you think about this implementation? I tried to keep the implementation general enough to accommodate all the use cases and to provide a first example of using errors with GraphQL schema.

xoldyckk:
One flaw is that it doesn't account for authorization logic for the user who's accessing that field. 

I think you're pushing way too much logic into the `createGraphQLConnection` function, it's job should only be to create and populate a connection object. Validation of args should be done within the resolver using seperate functions. The validator functions can also act as transformers transforming the arguments into things that mongoose accepts. For example returning `{ $gte: "id" }` or `{ $lte: "id" }` for arguments `after` and `before` respectively. Returning `limit` for `first` and `last`, and including the logic for `+1`, `+2` to the `limit`.

Your sorting logic is incorrect. Sorting order is determined by the client and it can be ascending/descending/default. A user cannot change sorting logic midway in a connection. A particular connection is heavily dependent on the sorting/filtering logic. A change to sorting/filtering logic midway in a connection won't work because position of edges is dependent on the sort/filter parameters. Any change in them should trigger a rebuilding of connection from the scratch.

Some things to note:-
1. Cursor is only there to uniquely identify an edge, it has no relation to sorting and filtering.
2. Sorting can exist independently without the cursor. For example fetching connection the first time without a cursor but providing a sort parameter.
2. Filtering can exist independently without the cursor. For example fetching connection the first time without a cursor but providing a filter parameter.

Uniquely identifying a field is a form of filtering as well, but in the code logic cursors can be better represented as just unique identifiers rather than filters.

I think the function can work for most of the connections. Sometimes though connection node might not be just a database object, it can be a custom object which exists only at the graphql layer, then this function won't work. But it usually never happens.

EshaanAgg:
- I don't think we need authorization logic to be present in the connectionFactory, as the same would be ensured by the resolvers of the particular field. 
- Totally agree on the points about pushing too much code into the `createGraphQLConnection` function, and would try to modularize it with the help of other helper functions.
- Got it. The sorting object basically needs to provide more filtering options, as the same should be chosen by client according to its particular use case. 
- It is true that this factory would only work for objects that can be fetched or created from the database, and would fail for virtual objects. But the same is a rare case (as you mentioned), and I can't seem to recall any instance of the same in the current API implementation.

Thank you for your time in reviewing this PR. I would get working on the changes you suggested.

EshaanAgg:
@xoldyckk Took your advice to separate `sorting` and `filtering` completely and abstracted away jobs to different functions. PTAL.

EshaanAgg:
@xoldyckk PTAL now. I have updated the connectionFactory and the resolver implementations in the hope of achieving the balance that you had described in your previous message. 
One change that I made was instead of allowing the client to send the `ConnetionArguments` type that you described in your comment, I defined an utility function that would do the said transformation for us. This allows us to get the best of both worlds, we allow the client to send arguments which are compliant with all the common GraphQL Cursor Pagination guidelines, and we also convert them to a form that allow our implementation to be simplified. What do you think about it?

src/typeDefs/errors/validatePaginationArgs.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const paginationErrors = gql`
+  union PaginationError =
+      PaginationArgsError
+    | MissingArguments
+    | IncorrectPairingOfArguments
+    | FetchLimitExceeded
+    | IncorrectCursor
+
+  interface PaginationArgsError {
```
xoldyckk, line 11: I think a much better interface would be this
```
interface Error {
  message: String!
  path: [String!]!
}
```

EshaanAgg: Got it!

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -36,3 +38,154 @@ export function graphqlConnectionFactory<T>(): InterfaceConnection<T> {
...
+  // If we are paginating backwards, then we must reverse the order of all fields that are being sorted by.
+  for (const [key, value] of Object.entries(sortingObject)) {
+    sortingObject[key] = value * -1;
+  }
+
+  return sortingObject;
+};
+
+// Generates the sorting arguments for filterQuery that can be passed into the .find() method
+export function getFilterQuery(args: ConnectionArguments) {
```
xoldyckk, line 95: Maybe name this one `getFilterObject`

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -36,3 +38,154 @@ export function graphqlConnectionFactory<T>(): InterfaceConnection<T> {
...
+  if (!allFetchedObjects || allFetchedObjects.length === 0)
+    return {
+      connectionData: connectionObject,
+      connectionErrors: null,
+    };
+
+  // Handling the case when the cursor is provided
+  if (args.cursor) {
+    // Validate the cursor
+    if (allFetchedObjects[0]._id.toString() !== args.cursor.toString()) {
```
xoldyckk, line 145: Same here, I think mongoose's `.equals()` method to compare objectIds is better and cleaner.

xoldyckk:
@kb-0311 What do you think of error types defined in this PR in graphql schema for connections? I think a controlled approach to this is needed. Some pre-defined error types which are prevalent across the schema so each resolver doesn't need to create its own set of error fields.

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -36,3 +38,154 @@ export function graphqlConnectionFactory<T>(): InterfaceConnection<T> {
...
+export const getLimit = (args: ConnectionArguments) => {
+  // We always fetch 1 object more than  args.limit
+  // so that we can use that to get the information about hasNextPage / hasPreviousPage
+  // When args.cursor is supplied, we fetch 1 more object so as validate the cursor as well
+  return args.cursor ? args.limit + 2 : args.limit + 1;
+};
+
+// Generates the sortingObject that can be passed in the .sort() method
+export const getSortingObject = (
+  args: ConnectionArguments,
```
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/utilities/graphqlConnectionFactory.ts
EshaanAgg: @xoldyckk 
I get your concerns, but I don't think it would be feasible to create utilities that can handle the complex nature of queries that have been highlighted in the later section of the attached link. It would make sense then that the developer himself writes the complete sorting and filtering logic himself.

Also currently, the filtering and sorting logic is strictly opt in only, as the `getSortingObject` function only return the `$lte` or `$gte` thingie if the `cursor` in supplied and the common practice of sorting by `_id` is used. If the same is not to be used, then these utilities can be ignored and a manual implementation can be done.

kb-0311:
@xoldyckk Yeah, some common error types prevalent across schemas can be defined as a collective, I agree.

I suppose you had a particular approach to that? Can we discuss some possible examples?

xoldyckk:
@kb-0311 That's for you to figure out. Some examples I can think of are these:-

1. Unauthenticated:- user should be signed in
2. Unauthorized:- authenticated user doesn't have permissions to do something
3. Forbidden:- something that is doable, but doesn't make sense according to real world logic, for example a user being a follower of themselves.
4. MinimumLength:- minimum length of a string
5. MaximumLength:- maximum length of a string
6. MinimumValue:- minimum value of a number(Int or Float)
7. MaximumValue:- maximum value of a number(Int or Float)
8. NotFound:- an item not found in database

One thing to note here is that not every error needs to be defined in schema. For example developer errors like fetching more items than the maximum limit can be prevented by client side developer, while they're developing the client app. They can hardcode the limit and make sure the limit is never exceeded. Same can't be said about end user of app, for example they may enter a firstname of 99999 characters which is non-sensical.

palisadoes:
@EshaanAgg any updates on this? The feature will be addressed in the UI redesign I'm sure

EshaanAgg:
@palisadoes Waiting for the PR to be reviewed. Merging this would complete the backend resolvers related to tags, and provide an easy way to work with connections for all developers.

Frontend changes related to tags would be addressed by the Admin UI Redesign project.

palisadoes:
Please work with your mentors on this. It's an important prerequisite addition

src/utilities/graphqlConnectionFactory.ts
EshaanAgg: @xoldyckk Can you please review?

xoldyckk: @EshaanAgg I thought we were waiting for standardization of errors. Without that I can't review as it'd be wasted effort.

palisadoes:
@EshaanAgg Any updates?

EshaanAgg:
@palisadoes I am currently waiting for the GraphQL error handling to be finalized a part of GSOC, and then I will complete this PR.

kb-0311:
@EshaanAgg For now, you can create a file and define some common errors mentioned in this comment https://github.com/PalisadoesFoundation/talawa-api/pull/1187#issuecomment-1529492528 for now in this PR. 
When I start working on the Error Handling part I will first explore all the resolver schemas, decide what is to be added, and then one by one make improvements to the common errors defined in that file.

If Adarsh is busy, you can ask a mentor to review it for now.

EshaanAgg:
Okay @kb-0311! Thanks. Will update the PR to reflect these changes by the end of this week.

EshaanAgg:
@xoldyckk Can you review now? I have added some base errors and tried to use them with pagination too. If you can review the PR once, then I'll work on adding the tests for the same!

xoldyckk:
ill take a look

EshaanAgg:
@xoldyckk Sorry for tagging you again, but any updates?

src/resolvers/UserTag/childTags.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (connectionErrors.length !== 0) {
+    return {
+      connectionData: null,
+      connectionErrors,
+    };
+  }
+
+  const newArgs = transformArguments(args);
+
+  const allUserObjects = await OrganizationTagUser.find({
```
xoldyckk, line 28: allUserObjects is incorrect name for the variable

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -36,3 +38,167 @@ export function graphqlConnectionFactory<T>(): InterfaceConnection<T> {
...
+  if (args.cursor) {
+    // Validate the cursor
+    if (allFetchedObjects[0]._id.toString() !== args.cursor) {
+      return {
+        connectionData: null,
+        connectionErrors: [
+          {
+            __typename: "IncorrectCursor",
+            message:
+              "The provided after cursor does not exist in the database.",
```
xoldyckk, line 160: remove `after` keyword from the message

src/typeDefs/errors/paginationError.ts
```diff
@@ -0,0 +1,24 @@
+import { gql } from "apollo-server-core";
+
+export const paginationError = gql`
```
xoldyckk, line 3: should be called `ConnectionError`

src/typeDefs/types.ts
```diff
@@ -385,13 +385,23 @@ export const types = gql`
...
+  }
+
+  type UsersConnectionResult {
+    connectionData: UsersConnection
+    connectionErrors: [PaginationError]
+  }
+
+  type UserTagsConnectionResult {
+    connectionData: UserTagsConnection
+    connectionErrors: [PaginationError]
```
xoldyckk, line 404: `[ConnectionError!]!`, make it a non null list of non null pagination errors, if no errors are found it'll be an empty list

xoldyckk:
@EshaanAgg Try to find out how mongoose sorting works if multiple fields are provided in the sort object. For example how it determines which field to sort by first and which after. If your implementation is correct then ignore this.

src/typeDefs/types.ts
```diff
@@ -385,13 +385,23 @@ export const types = gql`
...
     usersAssignedTo(
       after: String
       before: String
       first: PositiveInt
       last: PositiveInt
-    ): UsersConnection
+    ): UsersConnectionResult
+  }
+
+  type UsersConnectionResult {
```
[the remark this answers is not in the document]
EshaanAgg: But the obje

EshaanAgg:
Made all the relevant changes @xoldyckk.
The objects are sorted in the order of the properties provided. Cross-checked the implementation, and it seems to work fine. The same belief would be hardened after testing for all these files is added.
Is the current implementation good enough for me to start work on testing?

src/typeDefs/types.ts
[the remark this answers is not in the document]
xoldyckk: @EshaanAgg i meant replace `connectionData` with `connection` and `connectionErrors` with `errors`. They're inside a connection result type, so there's no point in prefixing their names with `connection`

xoldyckk:
I still haven't reviewed connection utilities thoroughly will do it soon.

src/typeDefs/types.ts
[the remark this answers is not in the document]
EshaanAgg: Got it!

EshaanAgg:
Okay @xoldyckk! Thanks a ton!

xoldyckk:
@EshaanAgg I think the relay connection arguments make the validation really verbose. We can instead use something like this:-

```graphql
# two possible directions to paginate from
enum PaginationDirection {
  BACKWARD
  FORWARD
}

interface CursorPaginationInput{
  # optional cursor because cursor isn't needed the first time a connection is triggered
  cursor: String
  direction: PaginationDirection!
}

# each connection should at the very least accept an optional cursor and a
# direction to paginate from as arguments
type UsersConnectionInput implements CursorPaginationInput{
  cursor: String
  direction: PaginationDirection!
  # optionally can provide sorting and filtering capabilities
}

type Query{
  users(input: UsersConnectionInput!): UsersConnectionResult!
}
```

Since we're not following the relay grapql spec, there's no need to adhere to it in an incomplete way.

EshaanAgg:
Yes @xoldyckk! Makes complete sense. It removes the complete overhead of actually transforming the arguments! I have updated the implementation to incorporate this schema.

palisadoes:
@DMills27 @CyberWake Please review too.

EshaanAgg:
@xoldyckk I had to change the schema a bit as we cannot use types as inputs according to GraphQL specs. Also, what do you think about the connection factory implementation? Is it good enough to start implementing tests for?

xoldyckk:
@EshaanAgg oh i see, inputs can't use interface type. That's kinda lame but whatever.

src/libraries/validators/validatePaginationArgs.ts
```diff
@@ -0,0 +1,25 @@
...
+  ConnectionError,
+  CursorPaginationInput,
+} from "../../types/generatedGraphQLTypes";
+
+export const validatePaginationArgs = (
+  args: CursorPaginationInput
+): ConnectionError[] => {
+  const connectionErrors: ConnectionError[] = [];
+
+  // Positive integer GraphQL Scalar ensures that the first and last are greater than zero
```
xoldyckk, line 12: remove this comment

src/libraries/validators/validatePaginationArgs.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Ensure that these arguments are less than the maximum allowed fetch limit
```
xoldyckk, line 13: fix the comment for limit

src/libraries/validators/validatePaginationArgs.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (args.limit > MAXIMUM_FETCH_LIMIT) {
+    connectionErrors.push({
+      __typename: "MaximumValueError",
+      message:
+        "More items than the allowed number of items were requested to be fetched.",
+      limit: MAXIMUM_FETCH_LIMIT,
+      path: ["limit"],
```
xoldyckk, line 20: ["input","limit"]

src/resolvers/UserTag/childTags.ts
```diff
@@ -0,0 +1,44 @@
...
+  })
+    .sort(
+      getSortingObject(args.input.direction, {
+        // The default sorting logic of ascending order by MongoID should always be provided
+        _id: 1,
+        name: 1,
+      })
+    )
+    .limit(getLimit(args.input))
+    .populate("userId")
```
xoldyckk, line 37: is populate needed here?

src/resolvers/UserTag/usersAssignedTo.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (errors.length !== 0) {
+    return {
+      data: null,
+      errors,
+    };
+  }
+
+  const allUserObjects = await TagUser.find({
+    tagId: parent._id,
+    ...getFilterObject(args.input),
```
xoldyckk, line 27: destructure before

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -36,3 +38,143 @@ export function graphqlConnectionFactory<T>(): InterfaceConnection<T> {
...
+  if (args.cursor) {
+    // Validate the cursor
+    if (allFetchedObjects[0]._id.toString() !== args.cursor) {
+      return {
+        data: null,
+        errors: [
+          {
+            __typename: "IncorrectCursor",
+            message: "The provided cursor does not exist in the database.",
+            path: [args.direction === "FORWARD" ? "after" : "before"],
```
xoldyckk, line 137: fix the path here

xoldyckk:
Connection factory function looks good enough yeah.

src/typeDefs/inputs.ts
```diff
@@ -244,6 +251,21 @@ export const inputs = gql`
     endTime: Time
   }
 
+  # Inputs in GraphQL cannot implement other inputs, but it must be ensured that connection inputs
```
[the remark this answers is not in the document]
EshaanAgg: Got it!

schema.graphql
```diff
@@ -279,6 +301,17 @@ input LoginInput {
 
 scalar Longitude
 
+type MaximumLengthError implements FieldError {
+  message: String!
+  path: [String!]!
+}
+
+type MaximumValueError implements FieldError {
+  limit: Int!
```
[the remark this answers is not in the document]
EshaanAgg: It can, but Field Error is too general a type. Maximum and Minimum value errors are good candidates to add more structure to our errors.

EshaanAgg:
@xoldyckk Made the requested changes and added testing for all the resolvers and functions! PTAL.

tests/libraries/validators/validatePaginationArgs.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    expect(payload.length).toBe(1);
+    expect(payload[0]).toMatchObject({
+      __typename: "MaximumValueError",
+      limit: MAXIMUM_FETCH_LIMIT,
+      path: ["input", "limit"],
+    });
+  });
+
+  it("returns empty error object if the args are valid", () => {
```
xoldyckk, line 23: error list

tests/resolvers/UserTag/childTags.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      },
+    };
+
+    const payload = (await childTagsResolver?.(
+      parent,
+      args,
+      {}
+    )) as UserTagsConnectionResult;
+
+    expect(payload.errors.length).toEqual(0);
```
xoldyckk, line 62: instead of testing everything individually, why not combine them into the shape of a connection object with expected values?

tests/resolvers/UserTag/childTags.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    const args: UserTagChildTagsArgs = {
+      input: {
+        limit: MAXIMUM_FETCH_LIMIT + 1,
+        direction: "FORWARD",
+      },
+    };
+
+    const payload = await childTagsResolver?.(parent, args, {});
+
+    expect(payload!.errors.length).toEqual(1);
```
xoldyckk, line 42: you have to be explicit and precise about the error that's going to be thrown, it cannot be disregarded as just an error, you at least need to check the`__typename`.

tests/utilities/graphqlConnectionFactory.spec.ts
```diff
@@ -1,12 +1,180 @@
...
+});
+
+describe("utilities -> graphqlConnectionFactory -> getLimit", () => {
+  it(`Should return 1 + limit if the cursor is not supplied`, () => {
+    const args: CursorPaginationInput = {
+      limit: 10,
+      direction: "FORWARD",
+    };
+
+    expect(getLimit(args)).toBe(11);
```
xoldyckk, line 37: Don't use hardcoded values, use the variables. Not only here, across all the places in other tests

tests/utilities/graphqlConnectionFactory.spec.ts
```diff
@@ -1,12 +1,180 @@
...
+
+    const payload = getSortingObject("FORWARD", sortingObject);
+
+    expect(payload).toEqual({
+      a: 1,
+      b: 1,
+    });
+  });
+
+  it(`Should return the supplied sorting object negated and in the same order when the direction is backward`, () => {
```
xoldyckk, line 66: i'm skeeptical about the order thing

tests/utilities/graphqlConnectionFactory.spec.ts
```diff
@@ -1,12 +1,180 @@
...
+
+    const payload = getFilterObject(args);
+
+    expect(payload).toEqual({
+      _id: { $lte: "12345" },
+    });
+  });
+});
+
+describe("utilities -> graphqlConnectionFactory -> getSortingObject", () => {
```
xoldyckk, line 122: This test suite is repeated, delete this one

tests/utilities/graphqlConnectionFactory.spec.ts
EshaanAgg: Got it. I am using hardcoded lengths in all other tests to manually ensure that the correct objects are fetched in correct order. Their data, though, is randomly generated every time.

tests/resolvers/UserTag/childTags.spec.ts
EshaanAgg: Sure will do that!

tests/utilities/graphqlConnectionFactory.spec.ts
EshaanAgg: I had tried using the same locally, and it seemed to work well for simple cases. For more complex cases with multiple fields, the implementation of the resolver would be custom, and thus this connection factor wouldn't be used. 

Is there anything you feel that needs to be added to the same?

EshaanAgg:
@xoldyckk Made the requested changes.

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -1,38 +1,178 @@
...
+      startCursor: null,
       endCursor: null,
       hasNextPage: false,
       hasPreviousPage: false,
-      startCursor: null,
     },
   };
 }
+// Generates the limit that can should be passed in the .limit() method
+export const getLimit = (args: CursorPaginationInput): number => {
```
xoldyckk, line 46: you can get away with passing just args.limit as the function argument

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -1,38 +1,178 @@
...
+  // We always fetch 1 object more than  args.limit
+  // so that we can use that to get the information about hasNextPage / hasPreviousPage
+  // When args.cursor is supplied, we fetch 1 more object so as validate the cursor as well
+  return args.cursor ? args.limit + 2 : args.limit + 1;
+};
+
+// Generates the sortingObject that can be passed in the .sort() method
+export const getSortingObject = (
+  direction: "FORWARD" | "BACKWARD",
+  sortingObject: { [key: string]: number }
```
xoldyckk, line 56: use Record<string, number> as sortingObject type

src/utilities/graphqlConnectionFactory.ts
```diff
@@ -1,38 +1,178 @@
...
+  if (!allFetchedObjects || allFetchedObjects.length === 0)
+    return {
+      data: connectionObject,
+      errors: [],
+    };
+
+  // Handling the case when the cursor is provided
+  if (args.cursor) {
+    // Validate the cursor
+    if (allFetchedObjects[0]._id.toString() !== args.cursor) {
```
xoldyckk, line 128: Is this a realistic check? If the cursor is provided and it is invalid, you either get a null or an empty list back which have already been tested against. If there's even one item in the list, that means cursor should be valid right?

palisadoes:
@DMills27 @CyberWake Please also comment on this

src/utilities/graphqlConnectionFactory.ts
EshaanAgg: Not necessarily. Since the check of the cursor is of `>=` or `<=` type, if I supply any cursor less than or greater than the valid cursor in both of the cases, respectively, the fetched results would be the same. A very simple cursor of `0` will be enough to hack the system, so to speak (though why any client may do so it something that's beyond me).

src/utilities/graphqlConnectionFactory.ts
EshaanAgg: I also need the cursor to check if I need to supply `limit + 1` or `limit + 2`

xoldyckk: my bad

src/utilities/graphqlConnectionFactory.ts
xoldyckk: Wait i didn't get that, what do you mean `0` formatted as mongoose object id and how does it relate to hacking the system?

EshaanAgg: Just simply meant that if we send 0 as a cursor in this implementation, it would be equivalent to sending to no cursor and all the results would be fetched.
Similary if lets say `abcd1` is a valid cursor, even `abcd2` would be treated as a valid cursor without this check (even though `abcd2` would have no object corresponding to it in the database)

xoldyckk: Since $gte is greater than equal, it won't pass if the objectId is not valid right?

EshaanAgg: It will. Lets say the objects are `abcd1`, `abcd3` and `abcd5`.
Then `abcd1` is a valid cursor, and would fetch all three of these objects, and then subsequently drop the first object as per this implementation and check.
If we are to remove it, then sending `abcd2` would also fetch the objects `abcd3` and `abcd5`, even though it isn't valid.

EshaanAgg: That would be a waste of a database request in my opnion. Why make a separate database request when we can validate the same in one?

xoldyckk: @EshaanAgg I'm not much aware with the advanced syntax of mongodb and mongoose, it could be that we can do both steps in a single query. I don't know. But yeah we have to use two queries in each of those resolvers. Again, the redundancy of doing the same thing can probably be abstracted away in a function or something.

EshaanAgg: @xoldyckk Got it. I have implemented the same in each resolver! PTAL.

src/resolvers/UserTag/childTags.ts
```diff
@@ -0,0 +1,61 @@
...
+    const cursorExists = await OrganizationTagUser.exists({
+      _id: args.input.cursor,
+    });
+
+    if (!cursorExists)
+      return {
+        data: null,
+        errors: [
+          {
+            __typename: "IncorrectCursor",
```
xoldyckk, line 35: InvalidCursor would be a much better term for this

tests/resolvers/UserTag/usersAssignedTo.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    const payload = await usersAssignedToResolver?.(parent, args, {});
+
+    expect(payload!.errors.length).toEqual(1);
+    expect(payload!.errors[0]).toMatchObject({
+      __typename: "IncorrectCursor",
+    });
+    expect(payload!.data).toBeNull();
+  });
+
+  it(`returns correct connection object when the arguments are correct`, async () => {
```
xoldyckk, line 69: Have u tested whether forward/backward paginating and default sorting/filtering are working correctly? Doing it for each resolver would be cumbersome, since testing the orm itself would defeat the purpose of using orm. I guess we have to trust the orm on this. Anyway it'd be good to verify all of this is working as expected.

This would require a pre defined list of objects, you have to arrange the objects in that list differently and compare the arrangement with what mognoose query returns for different filter/sorts.

tests/utilities/graphqlConnectionFactory.spec.ts
```diff
@@ -1,12 +1,127 @@
...
+});
+
+describe("utilities -> graphqlConnectionFactory -> getLimit", () => {
+  it(`Should return 1 + limit `, () => {
+    expect(getLimit(MAXIMUM_FETCH_LIMIT)).toBe(MAXIMUM_FETCH_LIMIT + 1);
+  });
+});
+
+describe("utilities -> graphqlConnectionFactory -> getSortingObject", () => {
+  it(`Should return the supplied sorting object when the direction is forward`, () => {
```
xoldyckk, line 38: `should return the default sorting object`

tests/resolvers/UserTag/usersAssignedTo.spec.ts
EshaanAgg: @xoldyckk I tested this locally on my system for a couple of user tags and it seemed to work well. Any possible flaws or edge cases can only be highlighted with the adoption of the same i.e. when we add more pagination-based fields and clients begin to use them. 
In such cases, debugging the connection factory would debug all the resolvers too in one short.

EshaanAgg:
The test coverage is failing as it is comparing this PR to the codebase to a commit that is over 3 months old (when the PR was first made). All the modified and created files in this PR have 100% coverage.

tests/resolvers/UserTag/usersAssignedTo.spec.ts
xoldyckk: @EshaanAgg i'm talking about the default filter and sort

EshaanAgg: Yes. Moongoose sorts them in the order they are specified, so just negating them worked well locally.

xoldyckk:
Looks good enough

xoldyckk (approved):

palisadoes:
@DMills27 @CyberWake Please review and approve.

DMills27 (approved):

EshaanAgg:
@palisadoes Please merge. Got it approved by Dominic sir.

palisadoes:
Try to get the tests fixed by the next PR so that we can maintain our coverage.

EshaanAgg:
> The test coverage is failing as it is comparing this PR to the codebase to a commit that is over 3 months old (when the PR was first made). All the modified and created files in this PR have 100% coverage.

@palisadoes As mentioned here, the test coverage in indeed 100% and there is nothing to fix. This discrepancy is due to the long nature of this PR (over 3 months).

PR #1189, opened 2023-03-16
Add `createUserTag` mutation [Fixes part of #1126]

EshaanAgg:
Feature

**Issue Number:**

Fixes part of #1126 

Yes

**Snapshots/Videos:**
NA

NA

**Summary**
Adds the source code and tests for `createUserTag` mutation.

No

Yes

### DIFF
--- a/src/resolvers/Mutation/createUserTag.ts
+++ b/src/resolvers/Mutation/createUserTag.ts
@@ -42,6 +42,23 @@ export const createUserTag: MutationResolvers["createUserTag"] = async (
     );
   }
 
+  // Check if the user has privileges to create the tag
+  const currentUserIsOrganizationAdmin = currentUser.adminFor.some(
+    (organizationId) =>
+      organizationId.toString() === args.input.organizationId.toString()
+  );
+
+  if (
+    !(currentUser!.userType === "SUPERADMIN") &&
+    !currentUserIsOrganizationAdmin
+  ) {
+    throw new errors.UnauthorizedError(
+      requestContext.translate(USER_NOT_AUTHORIZED_TO_CREATE_TAG.MESSAGE),
+      USER_NOT_AUTHORIZED_TO_CREATE_TAG.CODE,
+      USER_NOT_AUTHORIZED_TO_CREATE_TAG.PARAM
+    );
+  }
+
   // Additonal checks if the parent folder is provided
   if (args.input.parentTagId) {
     const parentTag = await OrganizationTagUser.findOne({
@@ -70,23 +87,6 @@ export const createUserTag: MutationResolvers["createUserTag"] = async (
     }
   }
 
-  // Check if the user has privileges to create the tag
-  const currentUserIsOrganizationAdmin = currentUser.adminFor.some(
-    (organizationId) =>
-      organizationId.toString() === args.input.organizationId.toString()
-  );
-
-  if (
-    !(currentUser!.userType === "SUPERADMIN") &&
-    !currentUserIsOrganizationAdmin
-  ) {
-    throw new errors.UnauthorizedError(
-      requestContext.translate(USER_NOT_AUTHORIZED_TO_CREATE_TAG.MESSAGE),
-      USER_NOT_AUTHORIZED_TO_CREATE_TAG.CODE,
-      USER_NOT_AUTHORIZED_TO_CREATE_TAG.PARAM
-    );
-  }
-
   // Check if another tag with the same name exists under the same parent tag
   const anotherTagExists = await OrganizationTagUser.exists({
     ...args.input,
@@ -101,7 +101,9 @@ export const createUserTag: MutationResolvers["createUserTag"] = async (
   }
 
   // Creates new tag and returns the same
-  return await OrganizationTagUser.create({
+  const newTag = await OrganizationTagUser.create({
     ...args.input,
   });
+
+  return newTag.toObject();
 };
--- a/src/resolvers/Mutation/createUserTag.ts
+++ b/src/resolvers/Mutation/createUserTag.ts
@@ -7,6 +7,7 @@ import {
   INCORRECT_TAG_INPUT,
   ORGANIZATION_NOT_FOUND_ERROR,
   TAG_NOT_FOUND,
+  TAG_ALREADY_EXISTS,
 } from "../../constants";
 
 export const createUserTag: MutationResolvers["createUserTag"] = async (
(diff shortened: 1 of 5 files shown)


xoldyckk:
@EshaanAgg 
Take a look:- https://thepalisadoes-dyb6419.slack.com/archives/CSWH4PN0M/p1679053124253489

Are you sure implementing these new resolvers with the older approach a good idea?

EshaanAgg:
Strong advocate for the approach you are suggesting here as the translation thing is completely broken, and unmaintainable in the long run. It should be the duty of the client to provide the translations of all the errors that are being generated. 
The same is true for the way we are handling errors to. Custom error objects are great, but they do not provide granular control over which field exactly is erroring out, and how should the client instruct the user to resolve them.

But at the same time I think, implementing either of these changes at this point should be avoided. It is simply due to consistency. If we are to devise a new strategy to handle errors and translation, the same should be done first with the approval of the majority of the community. After the same is done, we should write some at least basic migration guides from the old approach to the new approach, so that the same can be utilised by all the contributors to completely migrate the codebase.

Migration to new strategy for some newly added features, while not immediately acting on the older ones is a bad practice in my opinion. However this doesn't mean that we shouldn't be actively discussing the newer solutions and their merits, as after all, things are indeed broken right now on the API side.

palisadoes:
> @xoldyckk Strong advocate for the approach you are suggesting here as the translation thing is completely broken, and unmaintainable in the long run. It should be the duty of the client to provide the translations of all the errors that are being generated. The same is true for the way we are handling errors to. Custom error objects are great, but they do not provide granular control over which field exactly is erroring out, and how should the client instruct the user to resolve them.
> 
> But at the same time I think, implementing either of these changes at this point should be avoided. It is simply due to consistency. If we are to devise a new strategy to handle errors and translation, the same should be done first with the approval of the majority of the community. After the same is done, we should write some at least basic migration guides from the old approach to the new approach, so that the same can be utilised by all the contributors to completely migrate the codebase.
> 
> Migration to new strategy for some newly added features, while not immediately acting on the older ones is a bad practice in my opinion. However this doesn't mean that we shouldn't be actively discussing the newer solutions and their merits, as after all, things are indeed broken right now on the API side.

@xoldyckk Can you write a brief and definitive summary of how we should handle errors in the API so that we can add it to Talawa-Docs as a reference for future developers? That way it won't get lost. Send me the markdown and I'll add it to that repo.

xoldyckk (commented):
@EshaanAgg Add the logic so that two tags with same name cannot exist under the same `parentTagId`

src/resolvers/Mutation/createUserTag.ts
```diff
@@ -0,0 +1,107 @@
...
+      requestContext.translate(TAG_ALREADY_EXISTS.MESSAGE),
+      TAG_ALREADY_EXISTS.CODE,
+      TAG_ALREADY_EXISTS.PARAM
+    );
+  }
+
+  // Creates new tag and returns the same
+  return await OrganizationTagUser.create({
+    ...args.input,
+  });
```
xoldyckk, line 106: return `lean()` here

src/resolvers/Mutation/createUserTag.ts
```diff
@@ -0,0 +1,107 @@
...
+      throw new errors.NotFoundError(
+        requestContext.translate(INCORRECT_TAG_INPUT.MESSAGE),
+        INCORRECT_TAG_INPUT.CODE,
+        INCORRECT_TAG_INPUT.PARAM
+      );
+    }
+  }
+
+  // Check if the user has privileges to create the tag
+  const currentUserIsOrganizationAdmin = currentUser.adminFor.some(
```
xoldyckk, line 74: This authorization check should happen right after you confirm whether the organization exists.

EshaanAgg:
@xoldyckk Make the required changes!

tests/resolvers/Mutation/createUserTag.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+      await createUserTagResolver?.({}, args, context);
+    } catch (error: any) {
+      expect(spy).toBeCalledWith(ORGANIZATION_NOT_FOUND_ERROR.MESSAGE);
+      expect(error.message).toEqual(
+        `Translated ${ORGANIZATION_NOT_FOUND_ERROR.MESSAGE}`
+      );
+    }
+  });
+
+  it(`throws TAG_NOT_FOUND error if the parent tag doesn't exist`, async () => {
```
xoldyckk, line 119: Change the message to include something along the lines `if parentTagId is provided and is not null`.

When writing test messages, think of how easy you can make for someone reading the test to understand what the test is testing for. Testing is a kind of documentation for your code, when looking at tests the intent should be clear.

tests/resolvers/Mutation/createUserTag.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+      );
+
+      await createUserTagResolver?.({}, args, context);
+    } catch (error: any) {
+      expect(spy).toBeCalledWith(TAG_NOT_FOUND.MESSAGE);
+      expect(error.message).toEqual(`Translated ${TAG_NOT_FOUND.MESSAGE}`);
+    }
+  });
+
+  it(`throws INCORRECT_TAG_INPUT error if the parent tag doesn't belong to the provided organization`, async () => {
```
xoldyckk, line 149: same as above

tests/resolvers/Mutation/createUserTag.spec.ts
EshaanAgg: Got it.

tests/resolvers/Mutation/createUserTag.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+      await createUserTagResolver?.({}, args, context);
+    } catch (error: any) {
+      expect(spy).toBeCalledWith(ORGANIZATION_NOT_FOUND_ERROR.MESSAGE);
+      expect(error.message).toEqual(
+        `Translated ${ORGANIZATION_NOT_FOUND_ERROR.MESSAGE}`
+      );
+    }
+  });
+
+  it(`throws TAG_NOT_FOUND error if the parent tag is provided (not null) but the same doesn't exist in the organization of the tag being created`, async () => {
```
xoldyckk, line 119: Make it `parentTagId is provided(not null) but no tag exists with _id === args.input.parentTagId`

EshaanAgg: Got it.

xoldyckk (approved):

palisadoes (approved):

PR #1217, opened 2023-03-22
Fixed extension rules lint warnings

### LINKED ISSUE #1202: Fix Extension Rules lint warnings
**Describe the bug**
Currently, there are too many **Extension Rules** ESLint warnings in the **develop** branch. 

Most of them are:-
"@typescript-eslint/no-empty-function",
"@typescript-eslint/no-unused-vars",

**To Reproduce**

1. Clone the **talawa-api** repository.
2. Install the project dependencies. (npm i)
3. In .eslintrc.json, change **Extension Rules**, from line 57 to 58, from "off" to "warn", just like this:-
pic3

4. Run **npm run lint:check**

**Expected behavior**
Lint should report 0 Warnings and 0 Errors

**Actual behavior**
Lint now reports too many warnings.

**Additional details**
Multiple contributors can work together as there are too many warnings.

KrutikaBhatt:
This PR fixes the extension rules warnings arising due to rules like `no-empty-functions` and `no-unused-vars`

**Issue Number: #1202**

Fixes #1202 

**Did you add tests for your changes?** Not needed. The files were tested again after making changes

**If relevant, did you update the documentation?** No

**Does this PR introduce a breaking change?** No

**Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-api/blob/master/CONTRIBUTING.md)?** Yes


KrutikaBhatt:
I have resolved majority of warnings. There were certain files with resolver functions that uses `_context` parameter. A resolver can optionally accept four positional arguments: (parent, args, context, and info). Since `_context` was not used anywhere is the function, the `no-unused-vars` warnings used to rise.

For now, I have just added `/* eslint-disable */` on top of the files to suppress it without modifying .eslintrc.json file. One of a better effective way to tackle this is to add this rule in the file.
```
"no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
"@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
```
Now we should be able to add an underscore to the start of your variable declarations and it should avoid the error message. This changes are preferable but do let me your views on it and if I should modify it accordingly

EshaanAgg:
@KrutikaBhatt @tasneemkoushar 
I think you should go ahead and edit the ealintrc.json to supress the warnings for warnings for variables prefixed with _ as using comments to disbalabe linting is a bad practice in general.

What do you think?

KrutikaBhatt:
> @KrutikaBhatt @tasneemkoushar I think you should go ahead and edit the eslintrc.json to supress the warnings for warnings for variables prefixed with _ as using comments to disable linting is a bad practice in general.
> 
> Making this change would ensure better linting and maintainance of codr quality for future as well.
> 
> What do you think?

Exactly. It could help in maintenance of code in the future. I think I will make these changes. When the function or variables find its use ,`_` could be removed

palisadoes:
Please fix the errors

palisadoes:
@tasneemkoushar Please review

xoldyckk:
@KrutikaBhatt `_` prefix is a typescript convention for unused variables and should only be used when variables need to be defined but aren't being used. For example function paramters:-
```ts
function myFunc(a: string, b: string) {
  return b;
}
```
In this function argument a needs to be written out, even though it's not being used in the function. In such a case write it like `_a:string`:-
```ts
function myFunc(_a: string, b: string) {
  return b;
}
```
If typescript-eslint complains even after doing this, you're free to suppress it using custom rule.

KrutikaBhatt:
> @KrutikaBhatt `_` prefix is a typescript convention for unused variables and should only be used when variables need to be defined but aren't being used. For example function paramters:-
> 
> ```ts
> function myFunc(a: string, b: string) {
>   return b;
> }
> ```
> 
> In this function argument `a` needs to be written out, even though it's not being used in the function. In such a case write it like `_a:string`:-
> 
> ```ts
> function myFunc(_a: string, b: string) {
>   return b;
> }
> ```
> 
> If typescript-eslint complains even after doing this, you're free to suppress it using custom rule.

The typescript-eslint showed warnings for basic `_context`,etc . Thats why I just suppressed this using this rule here
```"@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }]```
This rule simply suppresses all the no-unused variable error warning with the variables starting with `_`

KrutikaBhatt:
Yes, its not being used currently but it could be used in future. This variable is used in most of the resolver APIs and I referred Graphql resolver docs - these  parameters `(parent, args, contextValue )` are positional parameters. So I didn't deleted or made changes to this.
image

KrutikaBhatt:
Ok understood, I will just remove the `_context` parameter in this case. Definetly, in future if needed it could be coded back

KrutikaBhatt:
The modifications are made accordingly

xoldyckk:
@KrutikaBhatt you need to remove the lint ignore rule related to your issue.

KrutikaBhatt:
Yes yes, I removed the int ignore rule from `eslint.json`. There are no changes in it

xoldyckk:
@KrutikaBhatt you didn't commit that `eslintrc.json` file.

KrutikaBhatt:
I pulled the latest code from develop branch and `eslintrc.json` is as it is with no modifications

xoldyckk:
@KrutikaBhatt in `eslintrc.json` you have to get rid of these rules:-
https://user-images.githubusercontent.com/58592306/226185951-f7bd9f2c-ae1d-4adf-9a5f-fe337ef757ea.PNG

They're only there so lint checks don't fail, those rules are not being enforced and are turned off. Your job was to fix all the places in code that those rules were complaining about. Once they are fixed, you get rid of those rules, and there should be no lint errors related to those rules in the code.

KrutikaBhatt:
Ok Ok cool, now I understood

xoldyckk:
@KrutikaBhatt you have to get rid of those rules, they are already enforced by `typescript-eslint` by default. They don't need to be manually added in `eslintrc.json`. The only reason they were added was to turn them off.

KrutikaBhatt:
oh god, I am really sorry now 😅

xoldyckk (approved):

PR #1255, opened 2023-04-04
Create Validators in Utilities

### LINKED ISSUE #1238: Bug Report: Create Validators in utilities
**Describe the bug**
To check if a user with matching `id` exists, the code has been repeated in many resolvers. Instead validator `getValidUserById` or `ValidateUserById` can be created in utilities, this will make our code base more modular.
A similar approach can be followed for Event and Organization.

**Expected behavior**
Validators in utilities

**Actual behavior**
Code repetition in resolvers.

**Additional details**

beingnoble03:

**Issue Number:**

Fixes #1238 

Yes

**Snapshots/Videos:**

**Summary**

No

**Other information**

Yes


beingnoble03:
@xoldyckk @palisadoes ptal.

Should I add changes in various mutation/query resolvers too? I didn't push because number of changes was pretty high.

beingnoble03:
I didn't create a common validation function because
- It would easier to validate individual type. For example, let's say, if we wanted to add certain checks just for valid user, it would be easier to update the individual validate utility `getValidUserById`. It would not be a good idea to add some code
```ts
if (object instanceOf Interface_User) {
// additional check here
}
```
- Better readability of the individual functions, when using in resolvers.

beingnoble03:
@xoldyckk @palisadoes ptal.

Should I add changes in various mutation/query resolvers too? I didn't push because number of changes was pretty high.

EshaanAgg:
@beingnoble03 @xoldyckk 
Using a custom validator function does not make sense in my opinion. As evident from this PR, the number of changes for the same is too high, and the functional gain is practically none as we are just abstracting one database call into a singular function. 

On the other hand the same would make much more sense if we had a general and customizable function that offers the developer to create multiple queries on the basis of inputs provided. 

For example, a starting implementation of the same can be 
```ts
export const validateDatabaseObject = async <T extends { _id: Types.ObjectId }>(
  objectId: string,
  model: Model<T>,
  getObject: boolean,
  error: string
): Promise<T | boolean> => {
  if (getObject) {
    const object = await model.findOne({
      _id: objectId,
    });

    if (!getObject) {
      // Add code to throw the relevant type of error with requestContext.translate()
    }

    return object!.toObject();
  } else {
    // Repeat the logic above but with exists
  }
};
```

What is model provides is utmost flexibility. Some sample calls to the function would be as follows:
```ts
// Get the organization
const organization = validateDatabaseObject(
  "test1234",
  Organization,
  true,
  ORGANIZATION_NOT_FOUND_ERROR
);

// Check if the user exists
validateDatabaseObject("test12345678", User, false, USER_NOT_FOUND_ERROR);
```

The arguments clearly dictate what the function should do. Abstarting away single calls into functions is not beneficial in the long run. Creating higher-order functions that perform related functionalities together does. Otherwise we would be just be moving code around the repository un-necessarily
.

EshaanAgg:
> I didn't create a common validation function because
> 
> * It would easier to validate individual type. For example, let's say, if we wanted to add certain checks just for valid user, it would be easier to update the individual validate utility `getValidUserById`. It would not be a good idea to add such code in the common function
> 
> ```ts
> if (object instanceOf Interface_User) {
> // additional check here
> }
> ```
> 
> * Better readability of the individual functions, when using in resolvers.

Can you provide of an example of the same being in resolvers? 

As far as my experience goes, the most duplicated code in the resolvers if these checking the existence of these database objects. Any how the validate functions introduced by you aren't performing any business logic (like checking if the user has access to the organization or privileges to perform the particular action) as these checks are being done in the resolver itself. How would custom existence functions provide any better functionality here?

beingnoble03:
> > I didn't create a common validation function because
> > 
> > * It would easier to validate individual type. For example, let's say, if we wanted to add certain checks just for valid user, it would be easier to update the individual validate utility `getValidUserById`. It would not be a good idea to add such code in the common function
> > 
> > ```ts
> > if (object instanceOf Interface_User) {
> > // additional check here
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Better readability of the individual functions, when using in resolvers.
> 
> Can you provide of an example of the same being in resolvers?
> 
> As far as my experience goes, the most duplicated code in the resolvers is checking the existence of these database objects. Anyhow the validate functions introduced aren't performing any business logic (like checking if the user has access to the organization or privileges to perform the particular action) as these checks are being done in the resolver itself. How would custom validator functions provide any better functionality here?

@EshaanAgg It's not about how it exists now. Maybe in future, we add some feature on Blocked User from Talawa. Now, we need to check that if he/she is blocked or not (which would be needed in each resolver). So instead of changing every resolver for checking if the user is blocked or not, we would simply change that particular validate utility (`validateUserById` in this case) and throw some custom error for the same.

So, according to me, creating individual validate functions will simplify any changes in the database structure in the future. Also, this makes our code more modular.

EshaanAgg:
> @EshaanAgg It's not about how it exists now. Maybe in future, we add some feature on Blocked User from Talawa. Now, we need to check that if he/she is blocked or not (which would be needed in each resolver). So instead of changing every resolver for checking if the user is blocked or not, we would simply change that particular validate utility (`validateUserById` in this case) and throw some custom error for the same.

That is a different use case all together. Since the same would be specific to `User` and `Chat` models, so you will define a custom validator like the following: 
```
const userCanAccessChat = (userId, chatId) => {
    const user = await validateDatabaseObject(userId, User, true, USER_NOT_FOUND_ERROR);
    const chat = await validateDatabaseObject(
      chatId,
      Chat,
      true,
      CHAT_NOT_FOUND_ERROR
    );

    // Add logic to check if the user is blocked and throw relevant error
}
```
and use it all the relevant resolvers.

beingnoble03:
> > @EshaanAgg It's not about how it exists now. Maybe in future, we add some feature on Blocked User from Talawa. Now, we need to check that if he/she is blocked or not (which would be needed in each resolver). So instead of changing every resolver for checking if the user is blocked or not, we would simply change that particular validate utility (`validateUserById` in this case) and throw some custom error for the same.
> 
> That is a different use case all together. Since the same would be specific to `User` and `Chat` models, so you will define a custom validator like the following:
> 
> ```
> const userCanAccessChat = (userId, chatId) => {
>     const user = await validateDatabaseObject(userId, User, true, USER_NOT_FOUND_ERROR);
>     const chat = await validateDatabaseObject(
>       chatId,
>       Chat,
>       true,
>       CHAT_NOT_FOUND_ERROR
>     );
> 
>     // Add logic to check if the user is blocked and throw relevant error
> }
> ```
> 
> and use it all the relevant resolvers.

@EshaanAgg  Sorry, I think you misunderstood. This is not what I wanted to say. Let's say there's a user whose misusing Talawa in some way, And SuperAdmin wants to block this user from using Talawa. This is the scenario I am talking about. It's not specific to chat.

EshaanAgg:
> @EshaanAgg Sorry, I think you misunderstood. This is not what I wanted to say. Let's say there's a user whose misusing Talawa in some way, And SuperAdmin wants to block this user from using Talawa. This is the scenario I am talking about. It's not specific to chat.

Got it. The number of such specific checks that need to be applied to a model across the whole backend would be significantly limited, and can be incorporated into the design later if the need arises. One way of the same can be using the model provided to check if the query is being executed for an User, and then perform the additional blocked check  in the `validateDatabaseObject` itself.

Otherwise the most relevant way to handle the same would be in the login resolver itself, preventing the user from authorizing itself.

beingnoble03:
> > @EshaanAgg Sorry, I think you misunderstood. This is not what I wanted to say. Let's say there's a user whose misusing Talawa in some way, And SuperAdmin wants to block this user from using Talawa. This is the scenario I am talking about. It's not specific to chat.
> 
> Got it. The number of such specific checks that need to be applied to a model across the whole backend would be significantly limited, and can be incorporated into the design later if the need arises. One way of the same can be using the model provided to check if the query is being executed for an User, and then perform the additional blocked check in the `validateDatabaseObject` itself.
> 
> Otherwise the most relevant way to handle the same would be in the login resolver itself, preventing the user from authorizing itself.

You can see here, it's possible that the number of errors for validating an object can be many. We don't know the number of errors we need to throw for each specific validation check in a validation utility (like `getValiduserById`). So, creating a common function can become complex thing to deal with in future.

beingnoble03:
Also, for the `login` resolver part, if you take a look at all the mutations, we are not always checking the `currentUser` is valid or not. So, this can't be assured with just `login` resolver, for every use case.

EshaanAgg:
Yet again, the same is not a design constraint. 
Let's say the number of global checks for a valid user increases by a magnitude, and becomes unmaintainable in the `validateDatabaseObject` function. Then we can simply shift to implementing a `getValidUser` validator which would use the `validateDatabaseObject` and other checks together, and then migrate all the user resolvers to the same.

The main point is, the choice of a custom validator for a model should be OPT-IN. Checking of the existence of database objects is too general an idea for us to require custom validators for each model.

beingnoble03:
@EshaanAgg I don't think so. As I said, the number of errors for a validation can be more than one.
So, I don't think it would be a good idea to pass error as an argument to call a common function `validateDatabaseObject`.
And, if we are not passing the error in the common function, we will just be combining different validate utilities in a single module. Not a good practice in my opinion.

Also, as you said we can create a different utility if the checks for a certain type of object increases. I don't think it would be a good practice to call the validator `getValidUser` for validating users and a different validator `validateDatabaseObject` for other type of objects [except user].

EshaanAgg:
I think the name `validateDatabaseObject` is thowing you off. What I'm saying consists of two parts:-

1. The changes that are being introduced in this PR which are dealing with `JUST` database existence of objects. We should have a general function for this use case instead of a custom function for each model, as all the functions that you have defined are practically doing the same thing, we just the model and error objects changed. Perhaps this function should be better called as `fetchFromDatabase`. There is no business logic involved in the same. The error can be passed in by user, or can be parsed by the function on the basis of the model, as the same is deterministic (XXXX_NOT_FOUND_ERROR)

2. The second part of the conversation deals with the future prospects, where we would be having custom validators for each model, that would be performing actually business logic checks on the different models. No doubt, in such a case each model should have its own validator, that would perform the checks specific to that model. This validator would itself make use of the `fetchFromDatabase` function defined above. This validator could throw a variety of errors, based on the logic performed in it. The same should be strictly an opt-in for each model, and shouldn't just be created because each model requires a database existence check.

Hope I'm able to clear my POV now.

beingnoble03:
@EshaanAgg I understand the approach you are proposing. I am not saying what you are trying to say is wrong.

But I have a different POV. I want to make the code more modular so that in future if there's any new feature we want to add that includes validation of object, it doesn't become a complex task. I want that validation of every type of object has a single source of validation, so that refactoring the code will be easier. 

In your approach, the validation of every type of object depends on a single function. So, if there's any new feature we want to add regarding validations, it would need us to create a new validation function for a single type of object taking use of the existing common single function (made earlier). This would then require us to change the validation function in each resolver, whereas in the current approach, it would just require us to update the existing validate utility.

Yes, I know the modular approach may seem like a big task as of now, but it may be fruitful in future.

I have given a read to [modular programming](https://www.tiny.cloud/blog/modular-programming-principle/). Please give it a read.

Even I was first considering going with a single common function, but this article changed my mind 😂

EshaanAgg:
Getting your POV too @beingnoble03. Will definitely read the article you mentioned!

xoldyckk:
@beingnoble03 

You're complicating it for no reason. I personally think resolvers the way they are clearly express the intent, they're not at all complex to understand. Besides the approach for first checking the existence of a database object then throwing an error corresponding to it is just there for convenience of throwing errors because custom parsing of database errors requires a bit more work. 

The best and performant approach would be to directly run the mutations and work on the database errors thrown by mongoose(orm). The errors will include the info that certain objects do not exist for the input you passed in the database request. Though for some things like relations between objects and authorization seperate database requests are required.

So, if you're gonna do something like this I'd say work on parsing those database errors and throwing corresponding errors. You can do this by wrapping the resolver code in try/catch blocks. The database error handlers you create can be used in a modular way across the resolvers.

palisadoes:

beingnoble03:
@xoldyckk Got it.

beingnoble03:
By the way @xoldyckk, are you suggesting to add something like this in models directory itself?
```ts
User.post('save', function(error, doc, next) {
  if (error.name === 'MongoServerError' && error.code === 11000) {
    next(new Error('Email already exists.'));
  } else {
    next();
  }
});
```

I have referred this code from https://mongoosejs.com/docs/middleware.html#error-handling-middleware

xoldyckk:
@beingnoble03 I'm saying mongoose is an orm, and an orm already provides enough abstraction over raw database queries. Abstracting away the orm into functions isn't something needed right now I feel. The code is repetitive and redundant sure, but it  is clear and makes sense when you look at it. It's not a scale right now where we need to think about deduplication.



> By the way @xoldyckk, are you suggesting to add something like this in models directory itself?
> 
> ```ts
> User.post('save', function(error, doc, next) {
>   if (error.name === 'MongoServerError' && error.code === 11000) {
>     next(new Error('Email already exists.'));
>   } else {
>     next();
>   }
> });
> ```
> 
> I have referred this code from https://mongoosejs.com/docs/middleware.html#error-handling-middleware

Yes, this is what I was referring to, but it needs to be flexible, composable into individual units, so that it can be used to handle mongoose errors in different resolvers and send back appropriate error response.

palisadoes:
Closing. No updates since April 22

PR #1263, opened 2023-04-06
code refactored for removeOrganization.ts

### LINKED ISSUE #1261: Minor changes in removeOrganization.ts to enhance the performance and efficiency
**Describe the bug**
The removeOrganization.ts file has many flaws which can cause slower performance like,

- The organization.posts.forEach loop uses async/await syntax to delete each post and comments associated with it one by one, which can result in slower performance. A better approach would be to use Promise.all to run all the delete operations concurrently.
- In the current implementation, when removing an organization, we are fetching each user that has joined the organization and removing the organization ID from their joinedOrganizations field one by one. This is not efficient for large organizations with many members. A better approach would be to use the $pull operator to remove the organization ID from all affected users' joinedOrganizations field in one go.
- The code has several nested if statements to handle errors and check user roles. This can make the code harder to read and follow. We can use early returns to exit the function early.
- The code currently calls the lean() method on several MongoDB queries. While this can improve performance, it can also cause issues with accessing related models or fields. Unless you're sure you don't need the functionality provided by the full model object, it's better to avoid calling lean().

**To Reproduce**

1. run the local server 'using npm run dev'
2. run the query removeOrganization

**Expected behavior**
The following changes can result in enhanced performance,

- Remove unnecessary lean() calls
- Use $pull operator to remove organization IDs from users' fields.
- Use Promise.all to delete posts and comments associated with the organization.

**Actual behavior**
The current implementation can can result in slow performance

**Screenshots**
image
image

**Additional details**
Not required

nishant3721:

Refactoring

**Issue Number:**

Fixes #1261 

No

**Snapshots/Videos:**

Not required

Not relevant

No

**Other information**

Not required

Yes


src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -27,7 +27,7 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
   async (_parent, args, context) => {
     const currentUser = await User.findOne({
```
EshaanAgg, line 28: Change the `findOne` to `exists` function as you aren't accessing any other properties on the `User` object in your resolver.

src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -54,100 +54,47 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
...
-      await Comment.deleteMany({
-        post: postId,
-      });
-    });
+    await Promise.all(
+      organization.posts.map(async (postId) => {
+        await Post.deleteOne({ _id: postId });
+        await Comment.deleteMany({ post: postId });
+      })
+    );
```
EshaanAgg, line 62: Use `deleteMany` function to perform these deletions.

```ts
await Post.deleteMany({
  _id: { $in : { organization.posts }}
})
```

src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -156,7 +103,7 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
     for (const blockedUserId of organization.blockedUsers) {
       const blockedUser = await User.findOne({
         _id: blockedUserId,
-      }).lean();
+      });
 
       await User.updateOne(
         {
```
EshaanAgg, line 109: Replace the same by a `pull` operation as used above.

src/resolvers/Mutation/removeOrganization.ts
nishant3721: ok

nishant3721: here we can't use exists() because we are also checking for super admin. Meaning, we have to pass the user info in it

src/resolvers/Mutation/removeOrganization.ts
nishant3721: then how we will delete the comments per post??

EshaanAgg: await Comment.deleteMany({
 post: { $in: organization.posts }
})

src/resolvers/Mutation/removeOrganization.ts
EshaanAgg: Got it. You can resolve this conversation.

EshaanAgg (commented):
LGTM. The assigned reviewers can comment better.

src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -27,7 +27,7 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
   async (_parent, args, context) => {
     const currentUser = await User.findOne({
       _id: context.userId,
-    }).lean();
```
xoldyckk, line 30: You talked about using `lean()` only when we're sure we're not going to use any of the other functionalities mongoose provides on the document, but I don't see you using any of those properties. What was it removed then?

src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -54,124 +54,52 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
...
-          $set: {
-            membershipRequests: requester?.membershipRequests.filter(
-              (request) =>
-                request.toString() !== membershipRequest?._id.toString()
-            ),
-          },
-        }
-      );
-    }
+    const membershipRequestIds = organization.membershipRequests;
```
xoldyckk, line 88: The variable isn't needed. If you want to fix the discrepancy of field names not having `Id` in the end of their names do it in mongoose models.

src/resolvers/Mutation/removeOrganization.ts
nishant3721: If we want to modify the data returned by the query, we should not use lean(). The returned object will not have the methods provided by Mongoose to modify and save the document.

nishant3721: Sir maybe in the future, we will use save(), virtual, getters,etc. According to my experience, I have not seen any query using lean(), If we want to modify the data returned by the query, we cannot use lean(). The returned object will not have the methods provided by Mongoose to modify and save the document.
But, I can revert this change sir. Waiting for your call on that!?

xoldyckk: @nishant3721 In the future if it's needed the simple fix is remove the `lean()` method, it's not that hard to do. Nowhere in the code you're using those mongoose document functions, so there's no point of removing the `lean()` method. It's put there in the first place because those things are not being used.

nishant3721: Sure I will revert this change in the pr

xoldyckk (approved):

palisadoes (approved):

src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -54,124 +54,55 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
...
-            ),
-          },
-        }
-      );
-    }
+    await MembershipRequest.deleteMany({
+      _id: { $in: organization.membershipRequests },
+    });
+    await User.updateMany(
+      { _id: { $in: organization.membershipRequests } },
```
xoldyckk, line 92: This is wrong.

nishant3721: Can you please elaborate a little bit sir. i was not able to get it!?

xoldyckk: @nishant3721 `organization.membershipRequests` is a list of `membershipRequestIds` not `userIds`.

src/resolvers/Mutation/removeOrganization.ts
```diff
@@ -54,124 +54,62 @@ export const removeOrganization: MutationResolvers["removeOrganization"] =
...
 
     /*
     Remove membershipRequest._id from each requester's membershipRequests
     field for membershipRequest.user for organization.membershipRequests list.
     */
-    for (const membershipRequestId of organization.membershipRequests) {
-      const membershipRequest = await MembershipRequest.findOneAndDelete({
-        _id: membershipRequestId,
-      }).lean();
+    const membershipRequest = await MembershipRequest.find({
```
xoldyckk, line 88: Rename this to `membershipRequests`

palisadoes (approved):

Issue #952, opened 2023-01-30
Refactoring the Image file utilities 


kb-0311:
**Describe the bug**
Currently the Image file handling utilities for the application are not the best approach to solve the problem. 

As Specified in this comment ->
https://github.com/PalisadoesFoundation/talawa-api/pull/948#issuecomment-1407703633

I think instead of whatever we are doing now the best possible approach would be to

1. Convert the client side image to binary string and then make a mutation using that uri as string to the API.
2. In the API , we update the Object (Post ,User , Organization) image field with that binary uri .
3. Whenever we need to fetch or render images we can just do it by converting the Buffer binary data string of the object image field.
This way -

1. We really reduce the size of the data because we are not sending and retrieving MBs of png files back and forth between API and client.

2. Fixes this legacy flaw that [this whole fiasco revealed.](https://github.com/PalisadoesFoundation/talawa-api/pull/948#issuecomment-1407699700)

I will need to change the following files ->
1. `src/utilities/uploadImage.ts`
2. `src/utilities/reuploadDuplicateCheck.ts.ts`
3. `src/utilities/deleteImage.ts`
4. `src/utilities/imageExtensionCheck.ts`
5. `src/utilities/imageAlreadyInDbCheck.ts`
6. `src/utilities/deleteDuplicatedImage.ts`
In the API project and also make changes for talawa-admin and talawa mobile app repo so that changes can be made to the client side also.
**Screenshots**

**Additional details**

kb-0311:
pls assign it to me

kb-0311:
[The issue I am also currently working ](https://github.com/PalisadoesFoundation/talawa-api/issues/887)on can only be merged after these changes are merged . Thus PR will need to wait until this one is done

I will take around 3-4 days to complete the work for this particular refactoring issue.
Until then I will complete one more small testing (non-utilities) issue

kb-0311:
@palisadoes pls checkout the comment under the now closed PR https://github.com/PalisadoesFoundation/talawa-api/pull/956#issuecomment-1410244587

kb-0311:
@palisadoes - Update:
I have been learning about how to achieve plugins in the API and admin console and what I have come up with is ->
1. I will create a  module for the Integration of S3 in its own separate modular class in its own directory under `src/CloudProviders/Amazon-s3` where I would create subdir and files for the complete integration of S3 in the app with #764 as reference.
[...]

PR #1268, opened 2023-04-07
Postimage

### LINKED ISSUE #1258: Add image upload in updatePost mutation
**Describe the bug**
Currently, there is no implementation of image upload in the updatePost mutation due to which the images are not getting updated.

Issue - https://github.com/PalisadoesFoundation/talawa-admin/issues/844

akankshat05:

Bugfix
**Issue Number:**

Fixes #1258 

Yes

**Snapshots/Videos:**

No

**Summary**

**Other information**

Screenshot_2

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -20,7 +20,7 @@ env:
           
 jobs:
   Code-Quality-Checks:
-    name: Check for linting, formatting and typpe errors
+    name: Check for linting, formatting and type errors
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository content
--- a/src/resolvers/Mutation/updatePost.ts
+++ b/src/resolvers/Mutation/updatePost.ts
@@ -8,6 +8,7 @@ import {
   LENGTH_VALIDATION_ERROR,
 } from "../../constants";
 import { isValidString } from "../../libraries/validators/validateString";
+import { uploadEncodedImage } from "../../utilities/encodedImageStorage/uploadEncodedImage";
 
 export const updatePost: MutationResolvers["updatePost"] = async (
   _parent,
@@ -71,14 +72,23 @@ export const updatePost: MutationResolvers["updatePost"] = async (
       LENGTH_VALIDATION_ERROR.CODE
     );
   }
-
+  let uploadImageFileName;
+  if (args.data?.imageUrl) {
+    uploadImageFileName = await uploadEncodedImage(
+      args.data.imageUrl,
+      post?.imageUrl
+    );
+  }
   return await Post.findOneAndUpdate(
     {
       _id: args.id,
     },
     // @ts-ignore
     {
-      ...args.data,
+      $set: {
+        ...args.data,
+        imageUrl: args.data?.imageUrl ? uploadImageFileName : post.imageUrl,
+      },
     },
     {
       new: true,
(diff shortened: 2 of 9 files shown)


xoldyckk:
@kb-0311 what do u think? If this keeps up all the implementation on client side for images uploads will be coupled to the hacky approach. 

What if they start implementing the video upload the same way. Should we reject the PRs?

kb-0311:
@xoldyckk @akankshat05 You can refer to this comment in the future for  general Issues regarding file uploads.

Yes, I do think generally at this time interval PRs regarding image/file uploads are generally pointless. unless they are discussing a long-term solution with a good high level approach taking in mind the differing file sizes and formats + multi tenant architecture + compression within the api server filesystem as @palisadoes had said that the files will need to be stored within the api server.

The reason for the hacky solution was an upsurge of client side cnotributors for outreachy suring which peter had asked me to atleast make the image uploads funcitonal by encoding so that the admin portal contributors have something to work with. Now that it is not the case I don't think we should rely on that encoding function anymore instead we should think of a more scalable long-term approach.

Why Encoding files is a bad idea->
https://github.com/PalisadoesFoundation/talawa-api/pull/956#issuecomment-1410244587

A better approach is to make use of `File` scalar in yoga or `Upload` scalar in apollo. Or only sending the metadata in graphql layer and actual content of the file through a separate service other than graphql layer(But I have not thought about this in detail). Why I could not do that at that time ->
https://github.com/PalisadoesFoundation/talawa-api/issues/952#issuecomment-1434149051


@akankshat05 Also I was thinking about implementing that function in updatePost too when I was implementing the entire encoded image thingy.
 Instagram does not allow any sort of updatetion on the image only caption and tags whereas Facebook does allow it. 
However, this is not the approach most platforms use for "updating" images in posts where uploading a new image directly overrides the previous one. You have to manually remove that picture (think of the small cross button on Slack/Facebook) and add a new one manually (unlike profile pics). And I am not even getting into multiple images yet.
This was my thought process for not implementing for updatePost whereas other mutations have it. It was not an overlooking mistake, it was because of business logic ambiguity.

palisadoes:
I asked about encoding as a hashing method to see how we could deduplicate uploaded files.

The intention wasn't to store the files as encoded data. I decided not to be more thorough in my questioning because I wanted a quick solution, the thread was becoming very long, and there was so much related Talawa activity at the time. We all knew it could have been done better, but expediency had to take priority. I should have emphasized hashing more too.

If there is a better simple intermediate fix, we should consider it. If I remember correctly, the current solution cannot handle video.

kb-0311:
@palisadoes Yeah, true. As I said A better solution for files would be a multipart form format which is the intermediate fix. Then comes the hashing and duplicate uploads checking 

>A better approach is to make use of File scalar in yoga or Upload scalar in apollo. Or only sending the metadata in graphql layer and actual content of the file through a separate service other than graphql layer(But I have not thought about this in detail). Why I could not do that at that time ->
https://github.com/PalisadoesFoundation/talawa-api/issues/952#issuecomment-1434149051

Ofc that can only be done after yoga migration which I was taking look into Eshaans branch but it will take a lot of consistent time and my end sem exams are starting again and will end on 10th may. So will go into that full throttle.

palisadoes:
Closing this and the issue until we have a permanent image upload solution.

PR #1300, opened 2023-04-14
Load Sample Data into MongoDB via a NPM Command

### LINKED ISSUE #1293: Load Sample Data into MongoDB via a NPM Command
Currently, whenever a contributor starts the server for the first time, they need to create user account, posts, events, etc by themselves.

Like we have a script to start the server ```npm run dev```

We can create another script ```npm run load:sample```

We need to define different sample JSON data of the collections, and import them using the mongo function.

A contributor changing the Database Collection names or adding some new fields would require to make changes in JSON data also, so imports can work in future too.

No Alternative currently

1. Create a script ```loadSampleData.ts``` 
2. Import JSON files and insertMany function will do the rest.
3. Create a command to trigger the script

**Additional context**
It would help in implementing tons of features in future, like on-the-fly development environments.

anshgoyalevil:
This PR adds a dynamic NPM command which parses the command line arguments and executes a data loader with those arguments.

**Issue Number:**

Fixes #1293 

No

**Snapshots/Videos:**
image
image

Yes

**Summary**
This PR adds a dynamic NPM command which parses the command line arguments and executes a data loader with those arguments.

No

**Other information**

## Syntax:

```npm run import:sample-data -- [args]```
You can pass the following arguments while running this script.
- ```--format```: Cleans the database before import. **Add this flag with caution. It will delete all of the existing data inside the talawa database.**
- ```--items=```: Specify the items to add.
    - Following ```items``` can be specified, separated with a comma ```,```
        - ```users```: For users collection
        - ```organizations```: For organizations collection
        - ```events```: For events collection
        - ```posts```: For posts collection
## Examples:
- ```npm run import:sample-data```: This command will import the complete sample database without removing the existing data.
- ```npm run import:sample-data -- --format```: This command will import the complete sample database after removing the existing data.
- ```npm run import:sample-data -- --format --items=users,organizations```: This command will import the sample ```users``` and ```organizations``` collections after cleaning the existing data.
- ```npm run import:sample-data --  --items=users,organizations```: This command will import the sample ```users``` and ```organizations``` collections without cleaning the existing data.

Yes

### CODE REVIEW

**Comment by anshgoyalevil** [comment]
@palisadoes 

Please review. I have updated the ```installation.md``` also.

For the automatic database cleanup and import, There are cloud functions available which trigger a script at regular interval of time.

For example, in Azure, we have [WebJobs](https://learn.microsoft.com/en-us/azure/app-service/webjobs-create)
In AWS, we have [AWS Lambda with CloudWatch](https://www.freecodecamp.org/news/using-lambda-functions-as-cronjobs/)

I am researching a better way for automation in cloud.

The

**Comment by EshaanAgg** [comment]
@anshgoyalevil Don't you think we should add a couple of more database objects to the samples?  One sample object per type itsn't really great for testing out the functionality of the app in my opinion.

**Comment by anshgoyalevil** [comment]
@EshaanAgg 

I was also thinking the same.

What do you suggest? Adding more users, orgs, or any other collection?

**Comment by anshgoyalevil** [comment]
Alright, thanks for suggestions.

File: src/utilities/loadSampleData.ts [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+// Required arguments that need to be parsed
+yargs.option("i", {
+  alias: "items",
+  describe: "Comma-separated list of collections to load sample data into",
+  type: "string",
+});
+
+yargs.option("f", {
+  alias: "format",
+  describe: "Whether to format the database before insertion",
```
EshaanAgg, line 17: `Formats all the collections present in the database before the insertion of objects. [WARNING] Use carefully.`

This is a better description is my opinion.

EshaanAgg:
@anshgoyalevil Do you think we should also have a confirmation of `Y/N` when the user uses the `format` flag? It would be a nice security feature in my opinion.

anshgoyalevil:
@EshaanAgg 

Yes that might have helped if we have a different command for format.

A user adding the ```format``` flag is already aware of the fact that database would be formatted.

package.json
```diff
@@ -37,6 +38,7 @@
   "homepage": "https://github.com/PalisadoesFoundation/talawa-api#readme",
   "dependencies": {
     "@graphql-tools/schema": "^8.4.0",
+    "@types/yargs": "^17.0.24",
```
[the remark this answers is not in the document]
anshgoyalevil: @EshaanAgg 

I think you haven't read the issue statement properly. This script not only needs to be executed in the dev environment, but required in production also.

It will be used by the CI/CD pipeline to execute and parse the arguments.

package.json
```diff
@@ -72,7 +74,8 @@
     "uuid": "^3.4.0",
     "validator": "^13.7.0",
     "winston": "^3.8.0",
-    "xss-clean": "^0.1.1"
+    "xss-clean": "^0.1.1",
+    "yargs": "^17.7.1"
```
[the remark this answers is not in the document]
anshgoyalevil: Same for this also.

package.json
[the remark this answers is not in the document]
EshaanAgg: Got it.

src/utilities/loadSampleData.ts
anshgoyalevil: Yes. Would change it.

vite.config.ts
```diff
@@ -30,6 +30,8 @@ export default defineConfig({
         "src/db.ts",
         "src/server.ts",
         "src/**/index.ts",
+        "src/utilities/sample_data/*",
+        "src/utilities/loadSampleData.ts",
```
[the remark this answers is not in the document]
anshgoyalevil: This is a CLI utility, and testing for it is redundant.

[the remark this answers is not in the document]
EshaanAgg: @anshgoyalevil 
It's not the CLI utility we should be testing but rather the logic of the functions that is included in the same. 
Like for example, is the `formatDatabase` function actually resetting the database or not, do the files for all the models that we are using to populate the database exist, etc.

The goal shouldn't be code coverage, but rather making sure that if in future if we add support for more models, to decide to rework the implementation of these functions, the same does not break the existing logic.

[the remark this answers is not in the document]
anshgoyalevil: @EshaanAgg 

Your concerns are valid, but take the example of Server.ts, are we testing it? No, because some functions and features shouldn't be tested. For example, we are creating an apollo server in server.ts, but would you go and test if the function is actually being returning a apollo server.

Similarly, we are working with FS in the loadSampleData.ts file.

We would need to mock fs, and yargs for the tests to work. We are acceesing the external functions which cannot be tested. 

I would write the test for it, but, I don't think it would help in any way.

anshgoyalevil:
> @palisadoes
> 
> I have updated the `installation.md` also.
> 
> For the automatic database cleanup and import, There are cloud functions available which trigger a script at regular interval of time.
> 
> For example, in Azure, we have [WebJobs](https://learn.microsoft.com/en-us/azure/app-service/webjobs-create) In AWS, we have [AWS Lambda with CloudWatch](https://www.freecodecamp.org/news/using-lambda-functions-as-cronjobs/)

@palisadoes 

Please review.

INSTALLATION.md
```diff
@@ -406,6 +407,53 @@ The key generated in the previous step is in a format suitable for use in a mobi
...
+## Sample Data Overview:
+
+The sample data contains organizations, users, events and posts. Here are the details for each of organizations and user account.
+
+### User Accounts and Organizations:
+
+| Email                 | Password | User Type  | Joined Organization | Admin For           |
+|-----------------------|----------|------------|---------------------|---------------------|
+| testuser1@gmail.com   | Pass@123 | USER       | Test Organization 2 | None                |
+| testuser2@gmail.com   | Pass@123 | USER       | Test Organization 2 | None                |
```
palisadoes, line 444: Use `example.org` instead of `gmail.com` for purposes of copyright.

sample_data/events.json
```diff
@@ -0,0 +1,32 @@
+[
+  {
+    "_id": "6437907585008f171cf2994c",
+    "recurring": true,
+    "recurrance": "ONCE",
```
palisadoes, line 5: Make all the events recurring. That way people will always see the events in the default calendar views which would be the current day, month, week.

palisadoes (changes requested):
See comments.

Also add sample_data directory to the `.github/workflows` authorized files YAML file so that we are always alerted to changes in this critical directory.

anshgoyalevil:
Sure Sir. Will make the changes related to above conversations.

sample_data/users.json
```diff
@@ -0,0 +1,177 @@
...
+    "joinedOrganizations": ["6437904485008f171cf29924"],
+    "registeredEvents": [],
+    "eventAdmin": [],
+    "adminFor": ["6437904485008f171cf29924"],
+    "membershipRequests": [],
+    "organizationsBlockedBy": [],
+    "status": "ACTIVE",
+    "pluginCreationAllowed": true,
+    "adminApproved": true,
+    "firstName": "Test",
```
palisadoes, line 18: Please give all the users real sounding names. This will be required if we launch a live demo. We need to start thinking this way from early. There should be random first and last name generators online.

sample_data/organizations.json
```diff
@@ -0,0 +1,87 @@
...
+    "_id": "6437904485008f171cf29924",
+    "status": "ACTIVE",
+    "members": ["64378abd85008f171cf2990d"],
+    "admins": ["64378abd85008f171cf2990d"],
+    "groupChats": [],
+    "posts": [],
+    "pinnedPosts": [],
+    "membershipRequests": [],
+    "blockedUsers": [],
+    "name": "Test Organization 1",
```
palisadoes, line 12: 1. Give the organizations real sounding names, descriptions and locations.
2. Don't use religious names as that may connote bias in a sensitive area. Don't use Palisadoes either.
3. You can end the names with `Charities`, `Foundation` etc. to make them universal.

palisadoes: Do this for all organizations.

palisadoes (changes requested):
See additional comments

anshgoyalevil:
Sure Sir, will make the changes, according to what industry leaders use. Like, will implement what BreezeCHMS use when we try to test a demo of their platform.

anshgoyalevil:
@palisadoes 

Sir, I have made the changes.

Gave real names to Users and Organizations, also updated the location.

Changed ```gmail.com``` reference to ```example.com```.

For the events, there is no recurring logic in the API layer, so, I haven't changed that field in the database.

palisadoes (approved):

PR #1304, opened 2023-04-14
Remove all queries that exposed unfiltered data to client [Fixes #1303]

### LINKED ISSUE #1303: Bug Report: Deprecate all the queries exposing unfiltered database objects to the client
**Describe the bug**
Currently, we are exposing the following query to the clients, which returns all of the database objects of the corresponding type to the client. This is not a good practice, as this allows clients to get all the details about all users (even from different organizations). Moreover, they does not serve any practical purpose.

- `directChats`
- `directChatMessages`
- `events`
- `getDonations`
- `groupChatMessages`
- `groupChats`
- `groups`
- `posts`

**To Reproduce**
NA

**Expected behavior**
NA

**Actual behavior**
NA

**Screenshots**
NA

**Additional details**
A similar example of the same was the `comments` query which was recently depreacted. All of these queries should be removed as well, as in condition the client should be allowed to query for all the database objects without checks without additional checks.

EshaanAgg:
Feature

**Issue Number:**

Fixes #1303 

NA

**Snapshots/Videos:**
NA

NA

**Summary**
Remove the schema definitions, resolvers and tests for the eight queries mentioned in the linked issue. 

No

**Other information**
NA

Yes


anshgoyalevil:
@EshaanAgg 

I think, we should apply the ```@auth``` and ```@role``` directive to these queries, instead of removing them.

In future, or maybe implemented during the GSoC period, a SUPERADMIN can see all of the Posts/Events/Comments regardless of the organization.

It may also help in plotting the Charts on SUPERADMIN Dashboard, with information about all of the Posts/Events/Comments. 

If we segregate the data this way, it would take a lot of queries to fetch the information during the Business Analytics Dashboard operations.

What's your view on this.?

EshaanAgg:
As I see it, even if a super admin wants to see all the posts, events or comments, the same information would be practically useless for the superadmin without context telling him/her from which organization the data is coming from. The role of superadmin is to have uninterrupted access to all the data and actions, which we aren't hindering. Keeping a query exposing unfiltered data just for SuperAdmin is an un-neccesary overhead for us.

If in GSoC period, if we require all the database objects to show to the SuperAdmin, then these queries can be easily restored. Also in such a case, the queries would be need to be reimplemeted from scratch to support pagination, as for any community with multiple organizations, the volume of such data would be enormous. As far as the statistics are concern, separate resolvers would be implemented for the same that would fetch the database objects, process them to calculate the statistics, and then return just the statistics to the client. It does not make sense to send all this data over network layer to the client and then for the client to calculate the statistics. 

But we shouldn't be providing support for things in the API which are not being used currently or in the near future.

xoldyckk:
Another thing to note is that connections are only available on unique/distinguishable objects(nodes). Either you created seperate queries to resolve each of those objects or you use `node` query resolver concept which uses `relay global object identification` pattern, also called the `node interface` pattern.

EshaanAgg:
@xoldyckk Yes I completely agree. Those resolvers could be created on an as-needed basis later, as the move to the `Global Object Identification` pattern would be a big change to the API and require a lot of refactoring of the code to be done.

xoldyckk:
@EshaanAgg are these queries not being used on client side?

EshaanAgg:
@xoldyckk They aren't being consumed in the clients anywhere. I had cross-checked them once before opening the issue to make sure that nothing breaks on the client side because of this change.

EshaanAgg:
@xoldyckk They aren't being consumed in the clients anywhere. I had cross-checked them once before opening the issue to make sure that nothing breaks on the client side because of this change.

xoldyckk (approved):

palisadoes (approved):

PR #1305, opened 2023-04-15
Proper Implementation of `Comments` and `Post` relation [Fixes #1302]

### LINKED ISSUE #1302: Proper implementation of the `comment` and `post` relation
Currently, the comments for a post are stored on the post model itself with the help of embedding of documents. This is not ideal, and instead we should define a separate relational schema called `CommentPost` which would help to act as a SQL join between these two relations, and thus handle the data for the same in a more scalable manner. 

This approach is analogous to the approach we have followed for `Tags` in the recently implemented designs.

Described above

NA

1. Create a new model `CommentPost` which will clearly show the `one to many` relationship between the Post and the Comment model.
2. Deprecate the `comments` field on the Post model.
3. Add a field level resolver for the `comment` field on the Post model in schema.
4. Change all the relevant resolvers for queries and mutations that use the comments model to use the relational schema.
5. Uodate all the relevant tests

**Additional context**
NA

EshaanAgg:
Feature

**Issue Number:**
Fixes #1302 

Yes

**Snapshots/Videos:**
NA

NA

**Summary**
Migrated the business logic to use a relational schema for the `post` and `comment` schemas, instead of embedding the comments on a post object in Mongo. Updated all the relevant resolvers.

No

**Other information**
NA

Yes


src/models/CommentPost.ts
```diff
@@ -0,0 +1,40 @@
+import { Schema, model, PopulatedDoc, Types, Document, models } from "mongoose";
```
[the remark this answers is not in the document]
EshaanAgg: Got it. Would reimplement all the changes to reflect the same.

src/resolvers/Mutation/removeComment.ts
```diff
@@ -48,36 +50,33 @@ export const removeComment: MutationResolvers["removeComment"] = async (
     );
   }
 
-  // Checks whether currentUser with _id === context.userId is not the creator of comment.
-  if (comment.creator.toString() !== context.userId.toString()) {
+  // Checks whether currentUser with _id === context.userId is not the creator of comment or the user is a SUPERADMIN
+  if (
+    comment.creator.toString() !== context.userId.toString() &&
```
xoldyckk, line 55: Shouldn't the organization admin be able to delete comments on a post that corresponds to their organization?

EshaanAgg: Yes! Will update the resolver to allow the same.

src/resolvers/Mutation/removeComment.ts
```diff
@@ -48,38 +52,43 @@ export const removeComment: MutationResolvers["removeComment"] = async (
     );
   }
 
-  // Checks whether currentUser with _id === context.userId is not the creator of comment.
-  if (comment.creator.toString() !== context.userId.toString()) {
+  const isCurrentUserAdminOfOrganization = currentUser.adminFor.some(
+    (organization) =>
+      organization.toString() === comment.postId.organization.toString()
```
xoldyckk, line 57: @EshaanAgg mongoose has inbuilt method of comparing mongodb objectIds. The method exists on every mongoose objectId. The comparision is done like this
`objectId1.equals(objectId2)`. Read more about it, and I'd say use it throughout the codebase whenever a comparision for mongodb objectIds is encountered.

src/resolvers/Mutation/removeComment.ts
```diff
@@ -48,38 +52,43 @@ export const removeComment: MutationResolvers["removeComment"] = async (
...
         commentCount: -1,
       },
     }
   );
 
-  // Deletes the comment.
+  // Deletes the comment
   await Comment.deleteOne({
-    _id: comment._id,
+    _id: args.id,
```
xoldyckk, line 87: You should stick to a convention for this. Either use `args.id` or use the database object you fetched `object.id`. Makes the code cleaner.

src/resolvers/Post/comments.ts
```diff
@@ -0,0 +1,8 @@
+import { PostResolvers } from "../../types/generatedGraphQLTypes";
+import { Comment } from "../../models";
+
+export const comments: PostResolvers["comments"] = async (parent) => {
+  return await Comment.find({
+    postId: parent._id,
```
xoldyckk, line 6: You know something like this makes sense to be populated with additional data like `user` who created the comment. A comment will mostly always be fetched with information about the user who created it, so it makes sense to include the data along with the comment.

src/resolvers/Query/post.ts
```diff
@@ -11,12 +11,6 @@ import { POST_NOT_FOUND_ERROR } from "../../constants";
 export const post: QueryResolvers["post"] = async (_parent, args) => {
   const post = await Post.findOne({ _id: args.id })
     .populate("organization")
-    .populate({
-      path: "comments",
```
xoldyckk, line 15: You forgot to implement a resolver for `creator` of the comment.

src/resolvers/Mutation/removeComment.ts
EshaanAgg: Thanks for the advice!

src/resolvers/Post/index.ts
```diff
@@ -0,0 +1,8 @@
+import { PostResolvers } from "../../types/generatedGraphQLTypes";
+import { comments } from "./comments";
+import { creator } from "./creator";
+
+export const Post: PostResolvers = {
```
xoldyckk, line 5: This isn't wrong but I meant resolver for creator of comment not creator of post. One more thing to notice here is that post and comment whenever they're fetched will probably always need the creator object along with them. So, populating creator each time a comment is fetched might be a good choice.

EshaanAgg: Makes sense. Would implement a `creator` resolver for the `comment` as well.

src/models/Comment.ts
```diff
@@ -19,7 +19,7 @@ export interface InterfaceComment {
  * @param text - Text
  * @param createdAt - Date when the comment was created
  * @param creator - Creator of the comment
- * @param post - Post to which the comment has been made
+ * @param postId - The post on which the comment is created
```
xoldyckk, line 22: `Id of the post on which comment is created`

src/resolvers/Mutation/removeComment.ts
```diff
@@ -48,38 +52,42 @@ export const removeComment: MutationResolvers["removeComment"] = async (
...
-  // Checks whether currentUser with _id === context.userId is not the creator of comment.
-  if (comment.creator.toString() !== context.userId.toString()) {
+  const isCurrentUserAdminOfOrganization = currentUser.adminFor.some(
+    (organization) => organization.equals(comment.postId.organization)
+  );
+
+  // Checks whether currentUser with _id === context.userId has the authorization to delete the comment
+  if (
+    currentUser.userType !== "SUPERADMIN" &&
+    !isCurrentUserAdminOfOrganization &&
```
xoldyckk, line 62: These should be `||` OR conditions not `&&` AND.

src/resolvers/Mutation/removeComment.ts
```diff
@@ -48,38 +52,42 @@ export const removeComment: MutationResolvers["removeComment"] = async (
...
 
-  // Checks whether currentUser with _id === context.userId is not the creator of comment.
-  if (comment.creator.toString() !== context.userId.toString()) {
+  const isCurrentUserAdminOfOrganization = currentUser.adminFor.some(
+    (organization) => organization.equals(comment.postId.organization)
+  );
+
+  // Checks whether currentUser with _id === context.userId has the authorization to delete the comment
+  if (
+    currentUser.userType !== "SUPERADMIN" &&
```
xoldyckk, line 61: These should be `||` OR conditions not `&&` AND.

src/resolvers/Post/creator.ts
```diff
@@ -0,0 +1,8 @@
+import { PostResolvers } from "../../types/generatedGraphQLTypes";
+import { User } from "../../models";
+
+export const creator: PostResolvers["creator"] = async (parent) => {
+  return await User.find({
```
xoldyckk, line 5: `.find()` always returns an array, you have to use `findOne`, `findUnique` or something.

src/resolvers/Query/postsByOrganizationConnection.ts
```diff
@@ -33,13 +33,13 @@ export const postsByOrganizationConnection: QueryResolvers["postsByOrganizationC
         pagination: true,
         page: args.skip,
         limit: args.first,
-        populate: ["organization", "likedBy", "comments", "creator"],
+        populate: ["organization", "likedBy", "creator"],
```
xoldyckk, line 36: Since you've implemented `creator` for `post` as well, `creator` is no longer needed here.

src/resolvers/Mutation/removeComment.ts
EshaanAgg: Lets suppose the user is a SUPERADMIN. Then the conditional in the first line would evaluate to False, and we would break out of the if clause and thus wouldn't throw an error. If we were to use OR, then we would proceed to check for other conditions too. 

The same logic can be extended to the other conditions as well. We are using AND here as the error should only be thrown if all three conditions are evaluated to true, i.e. the user does not have any of the three listed authorizations.

xoldyckk (approved):

palisadoes (approved):

Issue #1201, opened 2023-03-19
Fix Typescript Rules lint warnings


tasneemkoushar:
**Describe the bug**
Currently, there are too many **Typescript Rules** ESLint warnings in the **develop** branch. 

Most of them are:-
"@typescript-eslint/ban-ts-comment",
"@typescript-eslint/no-explicit-any",
"@typescript-eslint/no-inferrable-types",
"@typescript-eslint/no-non-null-asserted-optional-chain",
"@typescript-eslint/no-non-null-assertion",
"@typescript-eslint/no-var-requires",

**To Reproduce**

1. Clone the **talawa-api** repository.
2. Install the project dependencies. (npm i)
3. In .eslintrc.json, change **Typescript Rule**, from line 49 to 54, from "off" to "warn", just like this:-
![pic2](https://user-images.githubusercontent.com/58592306/226185366-2143823c-1101-403f-a04e-f5080316cc02.PNG)

4. Run **npm run lint:check**

**Expected behavior**
Lint should report 0 Warnings and 0 Errors

**Actual behavior**
Lint now reports too many warnings.

**Additional details**
Multiple contributors can work together as there are too many warnings.

tasneemkoushar:
We can merge a contributor's pull request if they are able to resolve at least 10 warnings, but we won't mark this issue as resolved until all warnings have been fixed. This allows for the creation of pull requests by many users, which we will then merge one at a time. While the remedy is minor, it is possible that two persons worked on the same warning. In that case, whoever solves it first will have their modifications merged.

ujwalkumar1995:
Hey @tasneemkoushar 
Could you please assign me this issue?
I will try to resolve as much of the issues as I can.

tasneemkoushar:
> Hey @tasneemkoushar Could you please assign me this issue? I will try to resolve as much of the issues as I can.

@ujwalkumar1995 I have assigned this issue to you, please go ahead.

nishtha981:
Heyy
Can I be assigned this issue?

SiddheshKukade:
I’d like to get assigned to this issue.

SiddheshKukade:
Status: I currently have an exam going for this week. I'll get started on this in a few  days


Exam timetable floor reference 

[...]

Issue #1275, opened 2023-04-09
Improve and Standardize Automated Linting Tests


palisadoes:
**Is your feature request related to a problem? Please describe.**

1. We currently have automated linting included in our PR process. 
2. Unfortunately we keep having issues where tests have warnings that could have been prevented with better linting. 
3. We need to look at the eslint plugin that we use and determine which additional linting rules should be enforced and implement them as part of our PR process. 
4. This will need to be synchronized between Talawa-API and Talawa-Admin
     - https://github.com/PalisadoesFoundation/talawa-api/issues/1275
     - https://github.com/PalisadoesFoundation/talawa-admin/issues/861

The list of potential rules can be found here:

- https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin/docs/rules 

**Describe the solution you'd like**

See above

**Describe alternatives you've considered**

See above

**Approach to be followed (optional)**

See above

**Additional context**

See above

KrutikaBhatt:
Can I work on this issue ?

PR #1306, opened 2023-04-16
Fix typescript lint warnings

KrutikaBhatt:
**What kind of change does this PR introduce?** This PR fixes the typescript lint warnings. Initially, there were more than 1200+ warnings which were then reduced.

**Issue Number: #1201**

The PR helps for solving issue #1201. Please keep the issue open since the PR doesn't completely fix the issue.

**Did you add tests for your changes?** - Yes. Tested for existing test cases. 

**If relevant, did you update the documentation?** - No

**Summary**
The error fixes the following lint warnings:

1. ban-ts-comment
2. no-explicit-any
3. no-inferrable-types
4. no-non-null-asserted-optional-chain
5. no-non-null-assertion
6. no-var-requires

This fix was necessary to understand what all new linting rules should be added and studied

**Does this PR introduce a breaking change?**  - No

**Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-api/blob/master/CONTRIBUTING.md)?** - Yes


anshgoyalevil:
@KrutikaBhatt 

You may also mention the issue https://github.com/PalisadoesFoundation/talawa-api/issues/1275 if this PR resolves it completely or partly.

palisadoes:
There should also be an entry in the [.eslintrc.json](https://github.com/PalisadoesFoundation/talawa-api/blob/develop/.eslintrc.json) configuration file too. Correct? I don't see it.

anshgoyalevil:
@KrutikaBhatt 

The code coverage is effected across all the files, since you have added a new branch in the test code using ```??``` operator, which is not covered by the existing tests.

Is there any workaround, to remove the warning while restricting the use of ```??```?

KrutikaBhatt:
> There should also be an entry in the [.eslintrc.json](https://github.com/PalisadoesFoundation/talawa-api/blob/develop/.eslintrc.json) configuration file too. Correct? I don't see it.

This PR is just a fix for the existing typescript warnings. I needed to fix them and study them before adding more linting rules in `eslint.json`. I have coded for #1275 and I will soon raise a PR. But I will try to improve the coverage before it

KrutikaBhatt:
> @KrutikaBhatt
> 
> The code coverage is effected across all the files, since you have added a new branch in the code using `??` operator, which is not covered by the existing tests.
> 
> Is there any workaround, to remove the warning while restricting the use of `??`?
> 
> You may also remove the `|| null` type across all interfaces in tests/helper functions. It would help reduce the branches.

I did checked for a lot of alternatives but ultimately have to use the fix method mentioned here - https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-non-null-assertion.md. Moreover the `no-non-null-assertion` & `no-non-null-asserted-optional-chain` were a bit complementary to each other. I will definitely try to check for improving the code coverage.

anshgoyalevil:
> > @KrutikaBhatt
> > 
> > The code coverage is effected across all the files, since you have added a new branch in the code using `??` operator, which is not covered by the existing tests.
> > 
> > Is there any workaround, to remove the warning while restricting the use of `??`?
> > 
> > You may also remove the `|| null` type across all interfaces in tests/helper functions. It would help reduce the branches.
> 
> I did checked for a lot of alternatives but ultimately have to use the fix method mentioned here - https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-non-null-assertion.md. Moreover the `no-non-null-assertion` & `no-non-null-asserted-optional-chain` were a bit complementary to each other. I will definitely try to check for improving the code coverage. 


I also encountered the exact same problem with these two rules.

They contradict each other. We would ultimately have to remove ```|| null``` type from interfaces.

If that too doesn't solve the issue, additional tests using the null data would have to be created to cover those ```??``` branches.

KrutikaBhatt:
> 
> I also encountered the exact same problem with these two rules.
> 
> They contradict each other. We would ultimately have to remove `|| null` type from interfaces.
> 
> If that too doesn't solve the issue, additional tests using the null data would have to be created to cover those `??` branches.

I don't feel removing `|| null` type from interfaces is a good option. It is a huge change. I am actually adding some new test cases rn to cover for `??` branches as much as possible from my side. In the future, let's just say if the `|| null` is removed. The codes can be refactored easily

EshaanAgg:
> > @KrutikaBhatt Why are we making changes in the non-null assertions in all test files (like from `testUser!._id` to `testUser?._id`)? The not-null assertion is a more descriptive and better assertion in my opinion, as we can be assured that our helper functions would always return a valid database object and would populate the global variable with the same.
> 
> This was actually one of the major reasons why I have not updated codes with `!` at many places and only fixed warnings where necessary. The non-null assertion operator asserts to the system that the object is not null or undefined but it is often a sign that code is not fully type-safe. It just helps to better structure program logic. Moreover, in our case, we have testcases that handle the null/undefined objects and do not let them pass as params in the main function body. I felt using `?` as a better alternative to resolve `@typescript-eslint/no-non-null-assertion` and a efficient way without introducing new codes or branches

I agree that the presence of `!` at many places indicates that the code is not fully type-safe and thus should be avoided as far as possible. But in all of the test files, the use of the same is a conscious choice as it is the easier way to write tests. Using the `?` operator is not a good practice, as we want to convey to the developers that these fields would indeed always exist on the provided objects.

My personal recommendation would be to use `@typescript-eslint/no-non-null-assertion` rule, but exclude all the code from the `tests` directory from the same, so that we continue to use the `!` operator in the same as a conscious choice. What do you think about this approach?

src/resolvers/Mutation/assignUserTag.ts
```diff
@@ -74,7 +74,7 @@ export const assignUserTag: MutationResolvers["assignUserTag"] = async (
   const requestUserBelongsToTagOrganization =
     requestUser.joinedOrganizations.some(
       (organization) =>
-        organization.toString() === tag!.organizationId.toString()
+        organization.toString() === tag.organizationId.toString()
```
xoldyckk, line 77: @KrutikaBhatt mongoose has inbuilt method of comparing mongodb objectIds. The method exists on every mongoose objectId. Comparision is done like this
`objectId1.equals(objectId2)`. Read more about it, and I'd say use it throughout the codebase whenever a comparision for mongodb objectIds is encountered.

src/resolvers/Mutation/createDirectChat.ts
```diff
@@ -50,7 +50,7 @@ export const createDirectChat: MutationResolvers["createDirectChat"] = async (
   const usersInDirectChat = [];
 
   // Loops over each item in args.data.userIds list.
-  for await (const userId of args.data!.userIds) {
+  for await (const userId of args.data?.userIds ?? []) {
```
xoldyckk, line 53: Much of these type errors in mutations are because of an imperfect graphql schema. When you're mutating things, you don't expect a nullable object from client, you expect a required object that has mutation data. Otherwise calling a mutation resolver makes no sense.

src/constants.ts
```diff
@@ -302,9 +302,9 @@ export const EMAIL_ALREADY_EXISTS_ERROR = {
 
 export const BASE_URL = `http://localhost:${process.env.port || 4000}/`;
 
-export const ACCESS_TOKEN_SECRET = process.env.ACCESS_TOKEN_SECRET;
+export const ACCESS_TOKEN_SECRET = `${process.env.ACCESS_TOKEN_SECRET || ""}`;
```
xoldyckk, line 305: These environment variables should be parsed/validated/transformed/type-asserted booting up the server. Something like zod or json schema would be a good tool to do this.

xoldyckk (commented):
@KrutikaBhatt I've seen you've made use of nullish coalescing operator `??`. It's a good practice. Whereever possible you should be using it instead of `||` operator.

KrutikaBhatt:
Yes sir, will make those changes

src/resolvers/Mutation/assignUserTag.ts
KrutikaBhatt: Yes sir, I have made these changes and tested them. Will push the changes

src/resolvers/Mutation/createDirectChat.ts
KrutikaBhatt: I made some basic modifications in graphql scheme for `input` and `mutations`. Similar errors were in `updateUserPassword` as well. The args.data will strictly required newPassword, confirmPassword and oldPassword. So the args. data was made a non-nullable field. Similar changes are made

src/constants.ts
KrutikaBhatt: Using `zod` is a very brilliant idea to tackle this issue. We should use it to validate the process.env variables as well. I will give it a try

KrutikaBhatt:
@xoldyckk, Thank you for reviewing. I have added zod for validating the process.env variables. For that we need to install two additional packages - `zod` and `zod-error`. I have made almost all the changes mentioned by you. Please do have a look.

This is how the console logs under different situations while validating the env variables:
1. Missing env variables or wrong inputs : It logs the appropriate message and process stops
image

2. When everything is good to go!
image

KrutikaBhatt:
@palisadoes, I have resolved merge conflicts, please do review it

palisadoes:
Why haven't you updated the eslint config file?

KrutikaBhatt:
Issue #1201 had a lot of error messages. This PR just solves majority of them. It was essential to make some improvement before enforcing new linting rules. The new changes in `eslint.config` file are covered in a separate PR #1310. The errors resolved so far here were arising due to these rules below :-
```
"@typescript-eslint/ban-ts-comment",
"@typescript-eslint/no-explicit-any",
"@typescript-eslint/no-inferrable-types",
"@typescript-eslint/no-non-null-asserted-optional-chain",
"@typescript-eslint/no-non-null-assertion",
"@typescript-eslint/no-var-requires",
```

palisadoes (approved):

palisadoes:
I had to revert this. The API doesn't seem to be starting and our automated documentation update is failing.

- https://github.com/PalisadoesFoundation/talawa-api/actions/runs/4777180802/jobs/8493322352

palisadoes:
@KrutikaBhatt Please resubmit

anshgoyalevil:
@palisadoes 

Sir, I guess, the issue is with the Runner.

We have not given it appropriate time to start.

KrutikaBhatt:
> @KrutikaBhatt Please resubmit

Yes sir, I will check it out

anshgoyalevil:
image

The Server has not been started and the schema is fetched before that, so the sleep timer here needs to be more, like 30 seconds for example.

@EshaanAgg 

Am I correct, or the error is due to something else?

palisadoes:
Why would it suddenly need more time?

anshgoyalevil:
@palisadoes 

It shouldn't require more time. The issue seems to be with something else. 

@KrutikaBhatt is server running fine in local environment?

EshaanAgg:
You are correct @anshgoyalevil. The server hadn't started, and that's why the automatical documentation generation failed. 
Why the server suddenly is taking so long so start is something that needs to be researched as linting changes are completely independent from server actions.

Can we rerun the workflow? Maybe it's a one time runner issue?

anshgoyalevil:
@EshaanAgg 

Yes, but, I checked the previous actions. There are similar outputs in the GitHub runner https://github.com/PalisadoesFoundation/talawa-api/actions/runs/4775300198/jobs/8489477847#step:8:29

This issue seems to be with the code changes itself.

palisadoes:
The job was re-run twice with the same failing result

KrutikaBhatt:
There might be a chance the server didn't start because of `zod`. There is a part of the code I added that states that if env variables are not set, the process ends with a warning log. I will fix this up and see if that's the main issue

palisadoes:
This means we need a test in the PR workflow that verifies that the API can start.

That should be a simple issue to create and resolve based on the push workflow

KrutikaBhatt:
> This means we need a test in the PR workflow that verifies that the API can start.
> 
> That should be a simple issue to create and resolve based on the push workflow

Yes, we do need something like this so that this same problem do not arise later on

KrutikaBhatt:
> > This means we need a test in the PR workflow that verifies that the API can start.
> > That should be a simple issue to create and resolve based on the push workflow
> 
> Yes, we do need something like this so that this same problem do not arise later on. I have rectified the issue and will create a new PR. Apologize from my end.

Please do review the PR here - #1328

PR #1317, opened 2023-04-20
Feature: Private events visibilty by UserType

### LINKED ISSUE #354: Feature request: Private Events
The participants in all events are publicly visible to everyone. There needs to be options for more privacy.

1. The ability to make events private where only participants can see who is attending
1. The optional ability to make private events invisible to persons not attending

N/A

N/A

**Additional context**
Tied to Issue https://github.com/PalisadoesFoundation/talawa/issues/983

kayceeDev:
This PR Fixes:
- The ability to make events private where only participants can see who is attending
- The optional ability to make private events invisible to persons not attending

**Issue Number:**
Fixes #354 

In progress

**Snapshots/Videos:**
```JS
 query getEventByUserType ($userId: ID!, $orgId:ID!){
  eventsByUserType(userId: $userId,orgId:$orgId){
    _id
    title
    description
    isPublic
    admins{
      _id
    }
  }
}
```

N/A

**Summary**
The query takes in the signed-in `userId` and `orgId` which is passed through the client and returns the events based on `UserType`

- `SUPERADMIN` will see all public and private events
-  `ADMIN` will see all organization's public and private events.
- `USER` will see all Organization's public events and only private events they registered as an attendee.

No, We only need to change the query used by the client to use this current implementation.

**Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-admin/blob/master/CONTRIBUTING.md)?**
Yes

### DIFF
--- a/src/resolvers/Query/eventsByUserType.ts
+++ b/src/resolvers/Query/eventsByUserType.ts
@@ -1,0 +1,90 @@
+import type { QueryResolvers } from "../../types/generatedGraphQLTypes";
+import type { InterfaceEvent, InterfaceUserAttende } from "../../models";
+import { Event, Organization, User } from "../../models";
+import { STATUS_ACTIVE } from "../../constants";
+import { getSort } from "./helperFunctions/getSort";
+/**
+ * This query will fetch all events for the organization which have `ACTIVE` status from database.
+ * @param _parent-
+ * @param args - An object that contains `orderBy` to sort the object as specified and `id` of the Organization.
+ * @returns An `events` object that holds all events with `ACTIVE` status for the Organization.
+ */
+
+export const eventsByUserType: QueryResolvers["eventsByUserType"] = async (
+  _parent,
+  args
+) => {
+  const sort = getSort(args.orderBy);
+
+  const user = await User.findById({ _id: args.userId });
+  const organization = await Organization.findById({ _id: args.orgId });
+  const filteredEvents: InterfaceEvent[] = [];
+  if (user?.userType === "SUPERADMIN") {
+    const events = await Event.find({
+      organization: args.orgId,
+      status: "ACTIVE",
+    })
+      .sort(sort)
+      .populate("creator", "-password")
+      .populate("tasks")
+      .populate("admins", "-password")
+      .lean();
+
+    events.forEach((event) => {
+      event.registrants = event.registrants.filter(
+        (registrant: InterfaceUserAttende) =>
+          registrant.status === STATUS_ACTIVE
+      );
+    });
+    return events;
+  }
+
+  if (organization?.admins.includes(args.userId)) {
+    const events = await Event.find({
+      organization: args.orgId,
+      status: "ACTIVE",
+    })
+      .sort(sort)
+      .populate("creator", "-password")
+      .populate("tasks")
+      .populate("admins", "-password")
+      .lean();
+
+    events?.forEach((event) => {
+      event.registrants = event.registrants.filter(
+        (registrant: InterfaceUserAttende) =>
+          registrant.status === STATUS_ACTIVE
+      );
+    });
+    return events;
+  }
+  const events = await Event.find({
+    organization: args.orgId,
+    status: "ACTIVE",
+  })
+    .sort(sort)
+    .populate("creator", "-password")
+    .populate("tasks")
+    .populate("admins", "-password")
+    .lean();
+
+  events?.forEach((event) => {
+    if (event.isPublic) filteredEvents.push(event);
+    if (
+      !event.isPublic &&
+      event.registrants?.some(
+        (registrant: InterfaceUserAttende) =>
+          registrant.userId === String(args.userId)
+      )
+    ) {
+      filteredEvents.push(event);
+    }
+  });
+
+  filteredEvents?.forEach((event) => {
+    event.registrants = event.registrants.filter(
+      (registrant: InterfaceUserAttende) => registrant.status === STATUS_ACTIVE
+    );
+  });
+  return filteredEvents;
+};
--- a/src/resolvers/Mutation/createMember.ts
+++ b/src/resolvers/Mutation/createMember.ts
@@ -1,4 +1,4 @@
-import { MutationResolvers } from "../../types/generatedGraphQLTypes";
+import type { MutationResolvers } from "../../types/generatedGraphQLTypes";
 import { errors, requestContext } from "../../libraries";
 import { User, Organization } from "../../models";
 import { superAdminCheck } from "../../utilities";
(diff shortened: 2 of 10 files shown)


palisadoes:
Please fix the failing tests. Our test code coverage system will fail if any of these conditions occur:

1) The overall code coverage drops below the target threshold of the repository
2) Any file in the pull request has code coverage levels below the repository threshold
3) The code does not comply with our linting check or other related formatting requirements 

We do this to ensure the overall reliability of the code base is maintained. This is a prerequisite for getting your PR approved. Assigned reviewers regularly review the PR queue and tend to focus on PRs that are passing.

palisadoes:
1. The expectation is that the SuperAdmin would access the Admin screens to do any per-organization activity.
2. The ability to see all activities for all organizations may be a feature added over the summer with GSoC. Until then this is the expected behavior.

kayceeDev:
Thanks @xoldyckk @palisadoes I will improve on your comments

kayceeDev:
> @kayceeDev Events are associated to organizations. On client side events are accessed using the organizationId. Even in the case of a superadmin it doesn't make sense to show them events from different organizations as it wouldn't make sense until events are associated to the organizations they're created for.
> 
> Even so, if superadmins need to see all events of all organizations query field just for that should be introduced. For organization admins and members, they access events on the organization object, as a nested relation.
> 
> If on client side, UI logic is implemented according to the changes you've made in this PR, I'd say the UI logic is completely wrong.

On the client side, the event I am filtering on is just organization events and not all events. Remember that we get the organization id from the query params and that's the id used for events for each organization. That's the event I'm filtering on the client as of now. Thanks for mentioning this. I just have to get organization events for the SUPERADMINS

xoldyckk:
@kayceeDev Something like filtering events by organizationId should never be done on the client side, it's the job of the server to provide that functionality. Also, UI should always be designed to lazy load stuff, not everything should be fetched at once. So, fetching all events of all organizations then ordering them according to organizationId doesn't make sense. 

Here's a sensible flow. You should first fetch the list of organizations, then there should be an button to view events of the organizations, when you click that button events for that particular organization are loaded, either in a tab or a new page.

kayceeDev:
> @kayceeDev Something like filtering events by organizationId should never be done on the client side, it's the job of the server to provide that functionality and also representing the relationships in a way that never leads to clients thinking about the relationships themselves. Also, UI should always be designed to lazy load stuff, not everything should be fetched at once. So, fetching all events of all organizations then ordering them according to organizationId doesn't make sense. 
> 
> Here's a sensible flow. You should first fetch the list of organizations, then there should be an button to view events of the organizations, when you click that button events for that particular organization are loaded, either in a tab or a new page.

Thanks @xoldyckk I completely understand the server should handle it and that's why I'm implementing that on the server now to replace what I did on the client to improve reusability. Thanks so much for your suggestions. I am currently writing the test cases. I will improve other parts and will need your review as I go.

palisadoes:
@kayceeDev I've restarted the workflow. Are you still able to complete this?

kayceeDev:
> @kayceeDev I've restarted the workflow. Are you still able to complete this?

Yes this should be working fine. The last issue I sorted out was the test cases. The lint's fails from what I can see. I will fix that and push again

palisadoes:
Are you still working on this?

kayceeDev:
I will definitely make a push EOD today

xoldyckk:
@palisadoes I've already reviewed this a month ago, the approach is wrong, you can read my messages above. If you still want to go with this approach then sure I can review the code itself.

kayceeDev:
> @palisadoes I've already reviewed this a month ago, the approach is wrong, you can read my messages above. If you still want to go with this approach then sure lemme know and I'll review the code.

Hello @xoldyckk can you elaborate on what you mean by the wrong approach? I have actually implemented what I thought you meant by the right approach. Can you tell me your thoughts process regarding this issue. If it's not what I already did. I will adjust the solution

xoldyckk:
@kayceeDev read the comments again, tldr is that this PR is trying to implement a feature which is a bad design choice.

kayceeDev:
> @kayceeDev Events are associated to organizations. On client side events are accessed using the organizationId. Even in the case of a superadmin it doesn't make sense to show them events from different organizations as it wouldn't make sense until events are associated to the organizations they're created for.
> 
> Even so, if superadmins need to see all events of all organizations query field just for that should be introduced. For organization admins and members, they access events on the organization object, as a nested relation.
> 
> If on client side, UI logic is implemented according to the changes you've made in this PR, I'd say the UI logic is completely wrong.

I took your comment into consideration. I have changed the solution of you look closely.  No SUPERADMINS only see events associated to only an organization they are  looking at not for all organizations. So also the admins of the respective organization. The query takes in useID and organizationID and then returns events that is accessible by the user given their roles. You should take a look and see the flow of the new query eventByUserType..

kayceeDev:
If this is merged, I will just undo the client and use this new query on the client to get events given a user ID and organizationID. @xoldyckk

xoldyckk:
Oh okay then, I'll take a look

kayceeDev:
> Oh okay then, I'll take a look

Thank you..

src/resolvers/Query/eventsByUserType.ts
```diff
@@ -0,0 +1,90 @@
...
+    .sort(sort)
+    .populate("creator", "-password")
+    .populate("tasks")
+    .populate("admins", "-password")
+    .lean();
+
+  events?.forEach((event) => {
+    if (event.isPublic) filteredEvents.push(event);
+    if (
+      !event.isPublic &&
```
xoldyckk, line 74: These two if statements can be combined into one using OR statement.

src/resolvers/Query/eventsByUserType.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @returns An `events` object that holds all events with `ACTIVE` status for the Organization.
+ */
+
+export const eventsByUserType: QueryResolvers["eventsByUserType"] = async (
+  _parent,
+  args
+) => {
+  const sort = getSort(args.orderBy);
+
+  const user = await User.findById({ _id: args.userId });
```
xoldyckk, line 19: Make use of either `findOne` or `find` mongoose functions, `findById` is very niche and specific to `id` which is not really required. This is to ensure there's a certain level of consistency in using mongoose functions.

palisadoes:
@kayceeDev Please make the suggested changes

palisadoes:
Unassigning due to inactivity and/or no submitted PR. Others need to get a chance. Please let us know whether you want to be reassigned before someone else requests assignment or select another issue.

PR #1349, opened 2023-06-25
Improved Event Management: Relevant backend changes

EshaanAgg:
This PR addresses all the concerns in the backend as a part of the [Improved Event Management Project](https://github.com/orgs/PalisadoesFoundation/projects/19?pane=issue&itemId=30292419)

### Changes 
- [X] Migrate the `EventAttendee` schema in the backend
- [X] Implement all the `EventProject` resolvers as discussed with Peter on Slack
- [X] Migrate the `Tasks` schema and all the relevant resolvers
- [X] Introduce a new `Volunteers` table
- [X] Link the `Volunteers` table to the already existing schema
- [X] Implement basic resolvers for volunteers
- [X] Implement all the resolvers for volunteers
- [X] Add event attendee resolvers

All the files modified/added have 100% test coverage.


DMills27 (approved):
Looks good to me

src/typeDefs/mutations.ts
```diff
@@ -73,7 +79,7 @@ export const mutations = gql`
 
     createUserTag(input: CreateUserTagInput!): UserTag @auth
 
-    createTask(data: TaskInput, eventId: ID!): Task! @auth
+    createTask(data: TaskInput!, eventProjectId: ID!): Task! @auth
```
xoldyckk, line 82: data -> input, move eventProjectId inside TaskInput

src/typeDefs/mutations.ts
```diff
@@ -9,6 +9,8 @@ export const mutations = gql`
```
xoldyckk, line 9: Use a fixed approach for writing schema definitions. This will create uniformity across the schema and make it easier to traverse through it manually. Graphql has no convention or strict rules for schema design, it is the developer's job to make conventions and strict rules for writing graphql schema for their applications. Here I'm listing some good conventions/patterns for mutations in graphql schema design:-

1. First you need name of the types/objects the mutation is to be implemented on. Then create mutation names by prefixing the name of those types/objects with action words like create, add, update, delete, remove etc. 

2. Have each mutation accept a single argument named input. Since the input is abstracted away into its own type in the schema it is much more cleaner and much more flexible to extend if needed later on. Make the input for each mutation required, it doesn't make sense to trigger a mutation without the input object. In exceptional cases input might not be needed.

3. If needed create a custom mutation result type to communicate to the client result and errors in a proper, easy to understand format.

Fixing the already existing resolvers is not as feasible because they've already been implemented on the client side and you'd need to update client side code as well. But the new ones you're created here can be written from the start to follow a good convention.

EshaanAgg: I have tried to follow the same as far as possible. In all of the mutations I have added, I have made sure only to use a singular input type for all the inputs as far as possible. The only time I have used two inputs, are when a `id` and some `data` input are required for the same, as it makes sense to take them separately IMO as the `id` is used in resolver in multiple places for auth checks and fetching, while the data is usually destructed in the resolver to populate. 

If you find any resolvers that differ from this convention, you can let me know and I will fix them.

EshaanAgg:
@xoldyckk I'm aware of the same, and sorry for the trouble! But I had previously discussed this will my mentor @DMills27 sir, that I would be working on the backend changes in one go and raising only one PR for the same as all of these changes were highly dependent on each other, and the reviewal process can often be slow.

If it helps, I have tested almost all of these changes locally with the corresponding client-side implementations on my system. I would be working on migrating most of these resolvers to pagination-based resolvers in the upcoming time. I would highly appreciate it if you can review this PR or provide a general overview of the quality of the PR.

xoldyckk (commented):
Some resolvers are typed with `any` keyword, fix that. 

You've used plain list types instead of paginating them in graphql schema, if the data is supposed to be small then there's no need to implement pagination, if not paginate them.

I haven't checked the nitty gritty details in each resolver and the corresponding tests as I don't have much context of what changes this is supposed to bring to the application.

@EshaanAgg There's too many redundant checks inside resolvers for whether user is authenticated, whether they're normal users, admins, superadmins. You can remove the redundancy using resolvers composition. Read this:- 
https://the-guild.dev/graphql/tools/docs/resolvers-composition

EshaanAgg:
Thanks, @xoldyckk. Fixed the use of `any` in some resolvers, as you mentioned. 

Redundancy of checks is a big issue across all of the codebase and should be worked upon immediately. But since that's not the prime motive of this PR, I have taken the liberty to other a [new issue](https://github.com/PalisadoesFoundation/talawa-api/issues/1352) for the same. I hope that new contributors will try to take that up, or I can pick it up after completing my project. 

Most of the data that is being handled by these resolvers would be small and thus would not require pagination. There are some fields (like the list of attendees for an event) that require filtering and pagination to be implemented, but I would be making a new PR for them after the basic functionalities in established in the client to demo an MVP as soon as possible.

xoldyckk (approved):

PR #1371, opened 2023-08-04
added run dev server action to pull request workflow

### LINKED ISSUE #1343: `CI/CD` : Check for Starting a Server.

For PR workflow under talawa-api repository, there should be a job that checks whether the development server starts successfully or not.

A job that checks whether the apollo server can start or not.

**Additional context**

ishtails:
PR CI/CD workflow feature

**Issue Number:**
Fixes #1343

No.

**Snapshots/Videos:**
N/A

No

**Summary**

Added a final github action to run the dev server using "npm run dev" before finishing checks as suggested by issue #1343

No

**Other information**
NA

Yes

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -97,6 +97,9 @@ jobs:
           path: './coverage/lcov.info'
           min_coverage: 95.0
 
+      - name: Run dev server
+        run: npm run dev
+
   JSDocs:
     name: 'JSDocs comments and pipeline'
     runs-on: ubuntu-latest


ishtails:
I have added a job that runs the dev server after testing is finished. Is there anything that I am missing? if so, let me know and i will make the required changes

ishtails:
Can i get some help with the checks not passing? @palisadoes 

- first check seems to say that i am not authorized to make changes to workflow files. is that so?
- dev server run workflow fails due to some kind of env issue i am guessing. do i need to configure env beforehand?

palisadoes:
1. Yes, we flag  changes to workflow and other key files so that the approvers can be extra vigilant.
2. Yes, you'll need to setup a `.env` file environment variable. Try setting up the API on your local machine to understand all the steps required. This is specifically why we created this issue to make sure the API would run correctly after each PR.

ishtails:
I have setup up a separate job for running dev server with dummy variables that worked on local setup, so i am assuming it should work on the workflow as well. let me know if i need to change anything

.github/workflows/pull-request.yml
```diff
@@ -112,4 +112,67 @@ jobs:
...
+        uses: actions/setup-node@v3
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Sleep for 10s
+        uses: juliangruber/sleep-action@v1
+        with:
+          time: 10s
+
```
noman2002, line 176: why we need to wait for 10 seconds, is there anything that will break without it ? @ishtails

ishtails: I don't believe we need to wait, but this sleep command was there in the test job as well before running tests, so I put it here to be on the safe side, as I felt it might be there for some specific reason...

ishtails: I will remove the sleep command from the job then, I also believe it is not required.

That is a good idea, to write a test for it. However I don't have any experience writing tests as of now. Can we have it as a separate issue maybe? So that either someone else could write the test for it, or I could just do it in the future.

xoldyckk: @ishtails It seems there are some oddities about github actions that I wasn't aware about. The 10 second wait seems to be done for the server to successfully start because the workflow doesn't wait for the server to successfully start I believe? Something like that. 

Also an action just to check if the server successfully starts seems wasteful, it'd be better to have this as a part of some end to end test workflow. For example this action https://github.com/PalisadoesFoundation/talawa-api/blob/develop/.github/workflows/push-documentation.yml already makes use of `npm run dev`. Starting the server itself doesn't tell us anything about whether all the services of the server are working perfectly.

ishtails: I am unaware of any such issues. Let me know if I should put the sleep command there or not...
If someone could explain the need for it, it would be really nice.

The issue originally just stated that there should be a job to check if the dev server starts or not, shouldn't the check passing or failing mean that the server started successfully or failed to start anyway?

xoldyckk (approved):
Whether this job provides any value I'm not too sure. Anyway the configuration seems to be correct so it's okay to merge.

noman2002 (approved):
I also agree with @xoldyckk that if this add any value or not, but the changes looks good to me.

palisadoes:
@xoldyckk @noman2002 We created the issue because there were times when the PR tests passed, but the application would not start. Some contributors would make modifications to the code to to satisfy the code coverage requirements, but ignored the usability. This is to help prevent such short sighted work.

palisadoes:
This has to be reverted. There is a syntax error.

PR #1377, opened 2023-08-17
Added the Redis cache to Cache Organization objects

### LINKED ISSUE #1374: Redis Cache implementation (Improved Performance -GSoC)

For frequently accessed database collections Like `Posts`, `Comment`, `Organization`, `Event`, require a lot of round trips to the DB which take a lot of time.

To minimize the round trips to the database we will be using Redis cache to store the objects in memory of the API. Also there should be mechanism to update/evict the cache store at appropriate times to prevent stale data.

To Implement caching behavior for :
- [x]  Organization
- [ ]  Posts
- [ ]  Comments
- [ ]  Events 

**Additional context**

kb-0311:

**Issue Number:**

Fixes #1374 (part 1)

Yes
**Snapshots/Videos:**

**Summary**

Yes. I have added the doc for the same as well.

**Other information**

Yes.


kb-0311:
@kevoniat Pls do review

palisadoes:
> @Kevoniat Pls do review

@Rumone Please review too.

kb-0311:
The unauthorised access is due to the changes in push.yml and pull-request.yml files to add the redis service in our test container

src/resolvers/Mutation/acceptMembershipRequest.ts
```diff
@@ -93,9 +94,14 @@ export const acceptMembershipRequest: MutationResolvers["acceptMembershipRequest
         $pull: {
           membershipRequests: membershipRequest._id,
         },
+      },
+      {
+        new: true,
       }
     );
 
+    await cacheOrganizations([updatedOrganization!]);
```
Rumone, line 103: Do a null check in these instances instead of the "!" operator

src/resolvers/Query/organizations.ts
```diff
@@ -17,13 +19,29 @@ export const organizations: QueryResolvers["organizations"] = async (
...
+    console.time("redis");
+
+    const organizationFoundInCache = await findOrganizationsInCache([args.id]);
+    console.timeEnd("redis");
+
+    if (!organizationFoundInCache.includes(null)) {
+      return organizationFoundInCache;
+    }
+
+    console.time("db query");
```
Rumone, line 33: Remove console log

kb-0311:
@Rumone 

1. Removed the logs from the code. It was for a demonstration of the performance benefits which slipped in the commits. My bad.
2. Improved the Grammar of the section of the installation guide.
3. Wouldn't the null check within the `cacheOrganization` function take care of any runtime anomaly?I think the '!' was only added because the type of arguments for cache organization do not allow null in the array. But if at runtime an array containing null values is passed to the `cacheOrganziation` the pipeline won't execute it at all. The pipeline will only execute non-null elements of the array.


https://github.com/PalisadoesFoundation/talawa-api/blob/54437b0b566246473cf355863cd68c384044a237/src/services/OrganizationCache/cacheOrganizations.ts#L11-L18

Kevonia (changes requested):

Rumone:
> Wouldn't the null check within the cacheOrganization function take care of any runtime anomaly?I think the '!' was only added because the type of arguments for cache organization do not allow null in the array. But if at runtime an array containing null values is passed to the cacheOrganziation the pipeline won't execute it at all. The pipeline will only execute for non-null elements of the array.

Actually the '!' operator is essentially a promise to the compiler that the value will never be null. It would be better to put a null check conditional there and ignore potential null objects in the array.

src/constants.ts
```diff
@@ -393,3 +393,7 @@ export const SMTP_OPTIONS = {
   SMTP_PORT: process.env.SMTP_PORT,
   SMTP_SSL_TLS: process.env.SMTP_SSL_TLS,
 };
+
+export const REDIS_HOST = process.env.REDIS_HOST!;
+export const REDIS_PORT = Number(process.env.REDIS_PORT);
+export const REDIS_PASSWORD = process.env.REDIS_PASSWORD;
```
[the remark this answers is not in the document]
Rumone: I agree with the current implementation, this would create an unnecessary object

kb-0311:
@Rumone @Kevonia 
1. Increased the test coverage.
2. Added null checks before passing an object to `cacheOrganziation` 
3. Do let me know what to finally do with the Redis global variables.

Rumone (approved):

.github/workflows/push.yml
```diff
@@ -43,6 +52,8 @@ jobs:
           - 27017:27017
     env:
       MONGO_DB_URL: mongodb://localhost:27017/talawa-test-db
+      REDIS_HOST: localhost
```
Kevonia, line 55: why is these value hard-coded?

kb-0311: Ohhh no, that is not the config file that is the workflow file in CI/CD. Basically to test the application in the CI/CD, we have to provide the env variables for connection to a MongoDb database and redis instance. Otherwise the server will crash in the container.

I started the redis service inside the testing container and passed the env variables to connect to those services.

They need to be hard-coded because we have to pass the actual value for the connection to happen in the application within the container.

Kevoniat (approved):

PR #1384, opened 2023-09-07
User Update email already exists error fix

### LINKED ISSUE #1382: User Update fails if email not changed
**Describe the bug**
While editing User Details currently it is mandatory to change "Email" field to updated other fields. We should be able to update other fields (like first name, last name) without changing Email. 

1. Login to Talawa Admin Portal
2. Open any User details
3. Click on edit
4. Change any field except Email

**Expected behavior**
Other fields should get updated without changing Email.

**Screenshots**

https://github.com/PalisadoesFoundation/talawa-api/assets/123248648/e8df74da-d09d-4a1d-9b28-2899dc8e8223

nitishkumar333:
Bugfix

Fixes  #1382 

**Summary**
Added a condition that will check if the user changed the email or not. If the email is changed then only it will check if the email already exists or not. Also changed "find" to "findOne" as we know only one user will have a email and it is faster.

No

**Other information**

Yes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "license": "GNU General Public License v3.0",
       "dependencies": {
-        "@apollo/server": "^4.7.4",
+        "@apollo/server": "^4.9.3",
         "@graphql-inspector/cli": "^3.4.19",
         "@graphql-tools/schema": "^10.0.0",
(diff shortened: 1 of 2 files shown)


palisadoes:
@nitishkumar333 Please also add a test for this modification. We don't want the failure to happen again.

PR #1388, opened 2023-09-15
`Post` , `Comment` and `Event` Cache

### LINKED ISSUE #1374: Redis Cache implementation (Improved Performance -GSoC)

kb-0311:
This PR implements the Redis cache for the collections ->
1. `Post`
2. `Comment`
3. `Event`

Previously a PR to implement caching modules for Organization collections had been merged. This PR aims to do exactly that but for other collections where read operations will be significantly higher than write operations.

**Issue Number:**

Fixes #1374 

Yes.

**Snapshots/Videos:**

I will be making a commit within this PR itself as installation guide has to be updated for MacOS too.

**Summary**

**Other information**

Yes


kb-0311:
@rumone @kevoniat Please review.

tests/resolvers/Post/comments.spec.ts
```diff
@@ -28,9 +39,28 @@ afterAll(async () => {
...
+      {}
+    );
+
+    const comments = await Comment.find({
+      postId: testPost!._id,
+    }).lean();
+
+    expect(commentsPayload).toEqual(comments);
+  });
+  it(`returns the comment object for parent post from cache`, async () => {
```
Rumone, line 56: Both test cases seem similar how do you assert that the data is coming from cache?

src/services/redisCache.ts
```diff
@@ -0,0 +1,17 @@
+import { Redis } from "ioredis";
+import { REDIS_HOST, REDIS_PASSWORD, REDIS_PORT } from "../constants";
+
+const RedisCache = new Redis({
+  host: REDIS_HOST!,
+  port: REDIS_PORT || 6379,
+  password: REDIS_PASSWORD,
+});
+
+// Setting the limit of the max memory of the cache to 100MB
```
Rumone, line 10: What do you use to determine the required size of the cache or is this a standard size?

src/services/redisCache.ts
```diff
@@ -0,0 +1,17 @@
+import { Redis } from "ioredis";
+import { REDIS_HOST, REDIS_PASSWORD, REDIS_PORT } from "../constants";
+
+const RedisCache = new Redis({
+  host: REDIS_HOST!,
```
Rumone, line 5: Why is REDIS_HOST potential nullable?

src/services/PostCache/findPostsInCache.ts
```diff
@@ -0,0 +1,56 @@
...
+    try {
+      const postObj = JSON.parse(post);
+
+      // Note: While JSON parsing successfully restores the fields, including those with
+      // Mongoose Object IDs, these fields are returned as strings due to the serialization
+      // process. To ensure accurate data representation, we manually convert these string
+      // values back to their original Mongoose Object ID types before delivering them to
+      // the requesting resolver.
+
+      return {
```
Rumone, line 28: Not sure I understand by this was done like this

src/services/PostCache/findPostsInCache.ts
```diff
@@ -0,0 +1,56 @@
+import PostCache from "../redisCache";
+import type { InterfacePost } from "../../models";
+import { Types } from "mongoose";
+
+export async function findPostsInCache(
+  ids: string[] | any[]
```
Rumone, line 6: While is `ids` potentially an `any[]`?

src/services/PostCache/deletePostFromCache.ts
```diff
@@ -0,0 +1,9 @@
+import PostCache from "../redisCache";
+
+export async function deletePostFromCache(postId: string): Promise<void> {
+  const key = `post:${postId}`;
+
+  await PostCache.del(key);
+
+  console.log("Post deleted from cache");
```
Rumone, line 8: Remove console log. Do we have a dedicated logging tool for the API?

src/services/PostCache/cachePosts.ts
```diff
@@ -0,0 +1,23 @@
...
+        pipeline.set(key, JSON.stringify(post));
+        // SET the time to live for each of the organization in the cache to 300s.
+        pipeline.expire(key, 300);
+      }
+    });
+
+    // Execute the pipeline
+    await pipeline.exec();
+  } catch (error) {
+    console.log(error);
```
Rumone, line 21: Do we use console log for all our logging. If so then this is fine for now but we should look into a new logging tool such as Bunyan.

Rumone (changes requested):
All the changes look good and consistent excellent work 👍🏽

src/services/redisCache.ts
kb-0311: @Rumone Well The size of the cache for now is basically a placeholder value for now. In production we will obviously have to account for the size of the actual real data as well as the total size of the mainframe where the API is hosted since this is going to be an in-memory cache to determine the size of the Cache. 

a few months back @palisadoes had suggested that a 100GB VPS can be used for the API. So around 0.1 percent of that is 100MB and that is basically the size of the placeholder value

Do you think I should make the size a global variable so that it can be changed later?

tests/resolvers/Post/comments.spec.ts
kb-0311: Well, the assertion is a little naive but basically, we can say that the data is coming from cache from the code coverage of the file in the tests. 

If the returned data was from the db query (instead of the cahce) then that would mean the conditional statement to check whether the comment exists in the cache -> if true returning the given comment from cache would not be hit and that would have reflected in the code coverage.

src/services/PostCache/cachePosts.ts
kb-0311: Yeah Sorry, that's actually my bad, WIll make changes to remove console.log. 

I think we do use a logger library but I think that is more for logging errors for now.

src/services/PostCache/findPostsInCache.ts
kb-0311: Yeah so basically in the cache we are storing our data in JSON format as values, the problem is when we parse the JSON object after fetching a post, comment, org , event etc the values for fields after parsing an object are all converted to primitive data types. 

So in the resolvers/model there are certain types like Mongoose Object ID or JS date-time type which are all converted to string values. This was causing the tests to fail due to runtime errors and that's why this extra step was needed.

src/services/redisCache.ts
kb-0311: Ohh good catch , its not needed, will fix

src/services/PostCache/findPostsInCache.ts
kb-0311: the `any[]` type is also not needed, removed it.

src/services/redisCache.ts
Rumone: The value is used in one location so no need to make it a global variable.

src/services/PostCache/findPostsInCache.ts
Rumone: Noted

palisadoes:

Kevoniat (approved):

palisadoes:
@kb-0311 Can you improve the code coverage?

kb-0311:
@palisadoes The tests coverage reduction was due to the changes made by another PR. The points were resolved in a conversation in the Whatsapp group. @aashimawadhwa will be looking into it. The test coverage for the changes I am introducing in this PR is 100%.

palisadoes (approved):

palisadoes:
Thanks for the update

PR #1403, opened 2023-10-08
File getting deleted from FileSystem fixed

### LINKED ISSUE #1396: File gets deleted from FileSystem if it already exists
**Describe the bug**
If we upload a file which already exists in database as base64 encoded, then the file gets deleted from database as well as FileSystem. But the name of the file (reference) still remains in the database.

**To Reproduce**

1.  Upload a image in UserUpdate or Organization settings.
2.  Upload the same image again

**Expected behavior**
Image should not be deleted from FileSystem even if its uploaded again. Only the "numberOfUses" in mongobd should be increased by +1.

**Screenshots**

https://github.com/PalisadoesFoundation/talawa-api/assets/123248648/8f06d2c8-48cf-41ee-a764-f333c7989c08

nitishkumar333:
Bugfix

**Issue Number:**
Fixes #1396 

Yes

Yes


palisadoes:
@EshaanAgg @xoldyckk Pleases take a look at this

src/utilities/encodedImageStorage/uploadEncodedImage.ts
```diff
@@ -44,6 +40,10 @@ export const uploadEncodedImage = async (
     return encodedImageAlreadyExist.fileName;
   }
 
+  if (previousImagePath) {
+    await deletePreviousImage(previousImagePath);
+  }
+
```
EshaanAgg, line 46: Why are we deleting the image if it already exists?
Ideally, we should check here to see if the supplied image differs from the already stored image. If it is, we should delete the old image and add the new one. If not, we can do an early return.

src/utilities/encodedVideoStorage/uploadEncodedVideo.ts
```diff
@@ -45,6 +41,10 @@ export const uploadEncodedVideo = async (
     return encodedVideoAlreadyExist.fileName;
   }
 
+  if (previousVideoPath) {
+    await deletePreviousVideo(previousVideoPath);
+  }
+
```
EshaanAgg, line 47: Same as earlier.

src/utilities/encodedImageStorage/uploadEncodedImage.ts
nitishkumar333: > Why are we deleting the image if it already exists?

No, If the supplied image already exists then we are doing an early return with the path of the image.
If the supplied image does not already exists then we are deleting the previous image and updating with the supplied image.

> Ideally, we should check here to see if the supplied image differs from the already stored image. If it is, we should delete the old image and add the new one. If not, we can do an early return.

Yes you are right, and I think this code is doing the same as you mentioned.

```
const encodedImageAlreadyExist = await EncodedImage.findOne({
  content: encodedImageURL,
});

if (encodedImageAlreadyExist) {
  await EncodedImage.findOneAndUpdate(
    {
      content: encodedImageURL,
    },
    {
      $inc: {
        numberOfUses: 1,
      },
    }
  );
  return encodedImageAlreadyExist.fileName;
}
```

Please let me know if I am missing something.

palisadoes:
@kb-0311 do you have any comments on this PR?

kb-0311:
@palisadoes @nitishkumar333 Good catch. This bug can be fixed just by moving the previous image check after the condition to check if the image already exists, which is exactly what has been done here. Good Job.

@nitishkumar333 It would be helpful if you add a short video/screenshots to showcase that this functionality works exactly as planned after this fix.  Also please try to cover all lines of code changes. The test coverage has gone down significantly.

Other than that, LGTM, should work as expected.

palisadoes:
@nitishkumar333 please work on adding tests for your work. Do you need assistance in this process? If so, the others on this thread can help

nitishkumar333:
I think I found some more edge cases in this code. I am currently working to figure those out, will soon update about it. And also will be writing more tests.

kb-0311:
@nitishkumar333 I would advise you to make a doc or rather just write up all of the edge cases which are not being handled properly in a comment.

Also I noticed there has been an increase in lines of missed code can you please cover those as well?

nitishkumar333:
@kb-0311  After doing the change as in this PR, updating image is working as expected but in the database the "numberOfUses" increases by 1 everytime a existing image is uploaded.  (which should not be the case here as image is being used only once)

https://github.com/PalisadoesFoundation/talawa-api/assets/123248648/76a08b87-972a-4863-a221-6f20a00d64b0

Should I remove these lines to fix it ?

`await EncodedImage.findOneAndUpdate(
  {
    content: encodedImageURL,
  },
  {
    $inc: {
      numberOfUses: 1,
    },
  }
);`

kb-0311:
@nitishkumar333 So these are the following cases in case a duplicate image is uploaded->

1. the image uploaded is for a post but the same image is the display picture of a user / org (different objects are using the same binary content; number of uses for image = number of unique objects referencing it. Here instead of creating three files containing the same binary data, only one image file is created with a number of uses are 3, saving a lot of memory overhead) 
2. A user/org has a `previousImage` as the Display picture and during updation same image is uploaded as a replacement. (which is what you are talking about).

We will need to handle the edge case you are talking about without missing the 1st one.

I think we can just check whether the binary data exists in DB , if it does we can just check whether its reference is the same as the `previousImage`, if it is then we can do an early return there itself.

 
```javascript
  const encodedImageAlreadyExist = await EncodedImage.findOne({
    content: encodedImageURL,
  });

  

  if (encodedImageAlreadyExist) {
  
    if(encodedImageAlreadyExists.fileName===previousImage) {
        return encodedImageAlreadyExist.fileName;
    }
    await EncodedImage.findOneAndUpdate(
      {
        content: encodedImageURL,
      },
      {
        $inc: {
          numberOfUses: 1,
        },
      }
    );
    return encodedImageAlreadyExist.fileName;
  }
  
  if (previousImagePath) {
    await deletePreviousImage(previousImagePath);
  }

```

I think this code handles all of the cases mentioned.

nitishkumar333:
@kb-0311  Thanks for the detailed and clear explanation it really helped.

I have made the commit with the changes and also added tests for it. Not sure why coverage is still negative i think maybe because of other files ?

kb-0311:
@nitishkumar333 The indirect changes are because the test cases for other resolvers have not been added. Remember the test cases for utilities are different from the actual mutation input test cases .You can write the different image upload scenario test cases for the `updatePost` resolver. 

I think other than the indirect test coverage, lgtm.

palisadoes:
@kb-0311 @EshaanAgg do you have any more suggestions? Should we merge this in when the coverage for the patch is met?

kb-0311:
@palisadoes I think this can be merged after @nitishkumar333 makes a video demonstration as proof for all the test cases above for talawa-admin after that this PR can be merged. We don't need to over-scope PRs.Then later, another issue can be opened to fix the coverage as a good first issue for newer contributors. If the coverage remains broken for more than a week, I will fix it myself.

@nitishkumar333 please make a small video of talawa-admin showcasing that the image upload works as expected along with all the edge cases, after that this PR can be merged imo.

nitishkumar333:
@palisadoes  @kb-0311 I noticed that images do not get deleted when Posts are deleted. Should I create a new issue for it or commit it in this pr?

https://github.com/PalisadoesFoundation/talawa-api/assets/123248648/82c77af8-7b3e-438a-9549-356b0d3c320f

kb-0311:
@nitishkumar333 The same problem exists when an organization and/or a user are deleted ie `remoteOrganization` and `removeUser`(remove user function does not exist in the API currently, a resolver to completely delete a User from the database altogether might be necessary for the near future).

I think this should be a new Issue where you resolve this discrepancy for all three resolvers.

nitishkumar333:
@kb-0311 after I pulled the latest develop branch code coverage decreased drastically, is that a issue my side ?
sorry I am new to this

nitishkumar333:
video demonstration

https://github.com/PalisadoesFoundation/talawa-api/assets/123248648/68e04f6a-ae6e-46d8-bc5c-c10a8e78a0bd

I will open a new issue to increase the code coverage as soon as this pr this merged.

palisadoes:
@kb-0311 @EshaanAgg Please comment on this

nitishkumar333:
Hi, @kb-0311 please approve this PR or do let me know if any improvement is required.

I have written tests to cover indirect changes, will be opening a new issue for it after this one is approved.

kb-0311:
@palisadoes @nitishkumar333 LGTM, this can be merged.

EshaanAgg:
LGTM!

palisadoes (approved):

PR #1467, opened 2023-12-07
Added query comments by post

### LINKED ISSUE #1460: commentsByPost graphql query not present
**Describe the bug**
- Query called commentsByPost is not present in graphql queries which is expected from the talwa fetch comments service.

**To Reproduce**

-  https://youtu.be/86AzK_Y_GCQ?t=48

This issue is linked to [Issue 2190](https://github.com/PalisadoesFoundation/talawa/issues/2190)

**Expected behavior**
We need to have graphql query to fetch data as expected 

**Screenshots**
image

imshivam-gupta:

**Issue Number:**

Fixes #1460 

**Snapshots/Videos:**

image

No

**Summary**

Comments in app were not visible because it was expecting query commentsByPost which was not available. This PR added this feature so that all comments can be rendered for a particular post.

No

**Other information**
N/A

Yes

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ coverage
 node_modules
 src/types
 docs/Schema.md
+.eslintrc.json
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,3 +6,4 @@ node_modules
 src/types
 docs/Schema.md
 .eslintrc.json
+schema.graphql
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,6 +23,15 @@
       }
     },
     {
+      // graphql in root folder parses the schema.graphql file
+      "files": ["./schema.graphql"],
+      "parser": "@graphql-eslint/eslint-plugin",
+      "plugins": ["@graphql-eslint"],
+      "rules": {
+        "@typescript-eslint/naming-convention": "off"
+      }
+    },
+    {
       "files": ["tests/**/*"],
       "rules": {
         "no-restricted-imports": "off"
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,15 +23,6 @@
       }
     },
     {
-      // graphql in root folder parses the schema.graphql file
-      "files": ["./schema.graphql"],
-      "parser": "@graphql-eslint/eslint-plugin",
-      "plugins": ["@graphql-eslint"],
-      "rules": {
-        "@typescript-eslint/naming-convention": "off"
-      }
-    },
-    {
       "files": ["tests/**/*"],
       "rules": {
         "no-restricted-imports": "off"
--- a/src/resolvers/Query/commentsByPost.ts
+++ b/src/resolvers/Query/commentsByPost.ts
@@ -1,0 +1,17 @@
+import type { QueryResolvers } from "../../types/generatedGraphQLTypes";
+import { Comment } from "../../models";
+
+/**
+ * This query fetch the Comment as a transaction for an organization from database.
+ * @param _parent-
+ * @param args - An object that contains `orgId` of the Organization.
+ * @returns A `Comment` object.
+ */
+export const commentsByPost: QueryResolvers["commentsByPost"] = async (
+  _parent,
+  args
+) => {
+  return await Comment.find({
+    postId: args.id,
+  }).lean();
+};
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@ node_modules
 src/types
 docs/Schema.md
 .eslintignore
+schema.graphql
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,3 +5,4 @@ coverage
 node_modules
 src/types
 docs/Schema.md
+.eslintignore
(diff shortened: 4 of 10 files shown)


imshivam-gupta:
@noman2002 @palisadoes When i run npm run lint:fix i am getting no error

image

but in this bot I am getting error

image

Can you please suggest what to do here.

noman2002:
@imshivam-gupta Fix the failing test.

imshivam-gupta:
@noman2002 Can you please tell how to add .eslintrc.json linting options because it will require additional dependencies.

EshaanAgg:
@imshivam-gupta Instead of linting `schema.graphql`, you should configure it to be ignored by ESLint as it does not need to be linted as it is autogenerated.

imshivam-gupta:
Yes but after changing eslint graphql is linting fine but since I changed eslint then it is asking for lining the json file and that will be the case if I add graphql to ignore in eslint.

.eslintrc.json
```diff
@@ -22,6 +22,15 @@
...
     },
+    {
+      // graphql in root folder parses the schema.graphql file
+      "files": ["./schema.graphql"],
+      "parser": "@graphql-eslint/eslint-plugin",
+      "plugins": ["@graphql-eslint"],
+      "rules": {
+        "@typescript-eslint/naming-convention": "off"
+      }
+    },
```
EshaanAgg, line 33: Remove this rule.

src/resolvers/Query/commentsByPost.ts
EshaanAgg, line 1: Please add the tests for this as well.

EshaanAgg (changes requested):

imshivam-gupta:
@EshaanAgg I have added the test for ```commentsByPost``` and performed the above changes.

.eslintignore
```diff
@@ -4,4 +4,5 @@ build
 coverage
 node_modules
 src/types
-docs/Schema.md
\ No newline at end of file
+docs/Schema.md
+.eslintrc.json
```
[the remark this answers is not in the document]
imshivam-gupta: Done

EshaanAgg (approved):

imshivam-gupta:
@EshaanAgg I guess during last commit I by mistake put a backspace and that one space caused prettier to fail. I fixed it. Sorry for the inconvenience.

imshivam-gupta:
Typescript failure is there for node_modules which can't be corrected from my side. What to do now?

EshaanAgg:
@imshivam-gupta Pull the latest changes. I have corrected the workflow file.

EshaanAgg:
@imshivam-gupta You might need to pull the changes again. There was a typo, which I have now corrected.

imshivam-gupta:
@EshaanAgg these are my final changes. I have performed all checks locally also but still not getting why checking for unauthorized file changes fails

imshivam-gupta:
@EshaanAgg the files for which I wrote test are having 100% coverage but the coverage dropped in 1 of files which I didn't change at all.

EshaanAgg (approved):

palisadoes:
@xoldyckk Please review

xoldyckk:
The PR also doesn't account for pagination and simply exposes the complete list of comments on a particular post. Imagine the client app fetching `999999` comments at once, this would lead to a server/database crash.

It is also a graphql anti-pattern, because it doesn't make use of the relation between a `Post` and the `Comment` created on that post to query related data. Instead of simply exposing `Post.comments` on a `Post` type it exposes a `Query.commentsByPostId` on root `Query` type. It makes the usage of graphql pointless since it's doing the same thing a REST api would, exposing a particular entity on a particular endpoint, providing no way to fetch along with it other entities related to it.

I have talked about these points before as well, in [discussions](https://github.com/PalisadoesFoundation/talawa-api/discussions) tab.

imshivam-gupta:
The complete app logic is built upon rendering complete data at a time. If we want to update that we have to modify a major chunk of the code to support the request to the server on scroll so that server load is reduced but I created this issue to create a graphql query expected on server side

palisadoes:
@xoldyckk I agree, we need to reduce the long term technical debt. Over the past 12 months we have been working hard to create a sustainable code base.
@imshivam-gupta Please resubmit according to the recommendations of @xoldyckk

palisadoes (changes requested):
@xoldyckk I agree, we need to reduce the long term technical debt. Over the past 12 months we have been working hard to create a sustainable code base.
@imshivam-gupta Please resubmit according to the recommendations of @xoldyckk

imshivam-gupta:
Ok I will work on this.

imshivam-gupta:
@xoldyckk Can you please confirm if I should fetch all the comments with the post only and remove this query?

xoldyckk:
@imshivam-gupta Yes, and it should also be paginated using a graphql connection.

imshivam-gupta:
I have checked the existing GraphQL schema and confirmed that it already satisfies the requirements for commentsByPost. There was no need for a separate query. I opened this issue initially to create APIs for the frontend, but I now realize that updating the frontend queries suffices. Thanks @xoldyckk  for providing this insight; it was overlooked by me earlier. I also noticed that the Talawa-API has many redundant queries, for which I will definitely create issues in the future. This way, we can maintain sustainable code without the need for additional queries. @palisadoes I am closing this pull request and will chnage the frontend query directly to fetch the comments.

PR #1489, opened 2023-12-11
Implements GraphQL Error Handling: Resolver user

### LINKED ISSUE #1474: Feature Request: Implement GraphQL Error Handling: Resolver user
Parent: #1450 
Implement Error handling in user resolver as mentioned in #1450.

We can set the return type of user resolver to ```UserResult``` which will contain the ```data``` and ```errors``` . ```errors``` can be ```UserNotFoundError``` or ```UnauthenticatedError```. The ```UnauthenticatedError``` can come from authdirective. Then we can handle all these errors in the client side.

Sivam2313:

Feature

**Issue Number:**

Fixes #1474 

Yes

**Snapshots/Videos:**

image

image

**Summary**

Now the user resolver return the data and the error array separately. I have added a ```newAuthDirectiveTransformer``` file to handle all ```UnauthenticatedError``` seperately in every resolver.

Changes the Schema in User resolver.

**Other information**

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -153,7 +153,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: dmnemec/copy_file_to_another_repo_action@v1.1.1
       env:
-        API_TOKEN_GITHUB: ${{ secrets.TALAWA_DOCS_SYNC }}    
+        API_TOKEN_GITHUB: ${{ secrets.GH_TOKEN }}    
       with:
         source_file: 'talawa-api-docs/'
         destination_repo: 'PalisadoesFoundation/talawa-docs'


xoldyckk:
Graphql working group is working on a solution for colocated errors, it is discussed in this rfc:- https://github.com/graphql/graphql-wg/blob/main/rfcs/ClientControlledNullability.md

With this errors would be catched at the component level by referencing its associated graphql fragment and mapping it to the `path` field of the corresponding error in root errors list.

This makes errors as data approach kinda useless. Since the graphql clients will allow us to declaratively handle the errors instead of imperatively checking the root errors array.

The errors as data approach isn't prevalent in talawa-api right now, they're only implemented for `userTags` field. So, it'd be better to refrain from using them and getting rid of them from the codebase in a graceful way.

Instead of errors as data approach, errors should just be thrown in the resolvers like the plain old way. 

@Sivam2313

Sivam2313:
So we dont have to change anything in the api and have to define fragments in the frontend to handle errors at the component level.

PR #1542, opened 2023-12-23
Feature: Audit trail fields in Schema

### LINKED ISSUE #1479: Feature Request: Implement Audit Trail Fields in Schema
Currently, the schema doesn't have basic audit trail features

Adding `createdAt`, `updatedAt`, `createdBy`, and `updatedBy` fields to each collection in the database. The goal is to automatically capture and store information about when and by whom a record was created and last updated.

LakshyaSatpal:
Feature

**Issue Number:**

Fixes #1479

yes

**Summary**
Added `createdBy`, `updatedBy`, `createdAt`, `updatedAt` fields to the following models:
1. Advertisement
2. Comment
3. DirectChat
4. Event
5. EventProject
6. GroupChat
7. Message
8. Organization
8. Post
10. Task

Added just `createdAt`, `updatedAt` timestamp fields to the following models:
1. Checkin
2. DirectChatMessage
3. Donation
4. Feedback
5. File
6. MessageChat

Added resolvers for the createdBy and updatedBy fields, for models, which have a mutation defined to update.
For example, Organization, Post, Task, Event, these models have a corresponding updateOrganization, updatePost, etc. mutation defined.

**Does this PR introduce a breaking change?**|
Changed the `creator` field to `createdBy` field at many models. The clients will have to be changed accordingly, if they use `creator` field.


LakshyaSatpal:
@noman2002 @aashimawadhwa Can you please check this and get it reviewed?

LakshyaSatpal:
@xoldyckk @EshaanAgg could you please review this PR?

sumitra19jha:
@LakshyaSatpal , implementation of capturing only the latest update (updated by) in the schema: @palisadoes  mentioned in the issue that for audit purposes, we need to track updates, so I think multiple updates  would be required not just the latest. Consider either adding nested documents within the current collection or creating a separate collection for tracking these updates. Initially, as our change volume is expected to be small, it's feasible to start with nested documents. Later, if the volume increases significantly, we can migrate to a separate collection. This approach allows us to maintain a comprehensive audit trail from the outset.

sumitra19jha (commented):
In the code please make sure to update the updatedBy. Also if you have changed any existing variable name, you might need to right a migration script for that as well.

sumitra19jha:
Leave the Migration Script part.

LakshyaSatpal:
> @LakshyaSatpal , implementation of capturing only the latest update (updated by) in the schema: @palisadoes mentioned in the issue that for audit purposes, we need to track updates, so I think multiple updates would be required not just the latest. Consider either adding nested documents within the current collection or creating a separate collection for tracking these updates. Initially, as our change volume is expected to be small, it's feasible to start with nested documents. Later, if the volume increases significantly, we can migrate to a separate collection. This approach allows us to maintain a comprehensive audit trail from the outset.

I've added a TransactionLog collection in this PR #1603,  which stores all the database changes. I believe this will solve the purpose of tracking all updates

xoldyckk:
@palisadoes `updatedBy` seems like a useless field to me. It will only ever be useful to know the last user who updated a particular record. It won't tell us about all the updates that particular record has gone through throughout its history. Keeping track of this stuff is a job for meaningful database/server event logs, not something to be put in the database fields.

palisadoes:
1. True, but the free version of Mongo doesn't do the logging we desire. 
2. This way we'll be able to chronologically check

palisadoes:
@LakshyaSatpal Please propose a solution where we use the filesystem for the logging output instead of the filesystem based on this suggestion.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1603#pullrequestreview-1802895013

palisadoes (changes requested):
@LakshyaSatpal Please propose a solution where we use the filesystem instead based on this suggestion.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1603#pullrequestreview-1802895013

palisadoes:
There are hundreds of files to review. In which one is the code for:

1. logging to  disk?
1. ensuring the the the desired logging output file is:
    1. configured in setup.ts?
    1. documented in INSTALLATION.md? and 
    1. given as an example in `.env.sample`?

Or is this part of another PR with this one just focusing on the Schema?

LakshyaSatpal:
This PR in only for updating the schema. The logging to disk is in the other PR.

palisadoes:
@xoldyckk @sumitra19jha Please review

xoldyckk:
@LakshyaSatpal 

I don't see the usage of `updatedBy` field in [this PR](https://github.com/PalisadoesFoundation/talawa-api/pull/1603), the transaction logs make use of `context.userId` to track the user user who executed a mutation/query, again what's the use of `updatedBy` field?

LakshyaSatpal:
1. `updatedBy` will store the user who last updated it.
2. transaction logs will store every change (create, update or delete) that happened in the database, along with who did it. 

As stated here #1479, by @palisadoes, we need `updatedBy` to track the id for the user that updated it. I understand the fact that Transaction Logs will store the same thing, by keeping all the updates.

Tbh, I don't have a strong opinion about this, totally depends on the requirement.

palisadoes:
If we can improve the maintainability by logging `context.userId` and that this approach records the user updating the document without needing `updatedBy`, then that's what we should use. The aim is to get a meaningful logging system with optimal effort.

It's more than just meeting the requirement which may be valid for an outsourced project, but if you have a vested interest in the success of the project the aim will be to make it better overall for everyone.

LakshyaSatpal:
Okay. Will work on removing the `updatedBy` field from the schema. 

@palisadoes I apologize if you got me wrong, I meant I'm unsure about the requirement of Admins and DevOps teams since this feature will be generally used by them only.

palisadoes:
We are both trying to figure this out together. It's new for me too. We'll get something usable in a few days.

xoldyckk (commented):
`creatorId` is a database field, in the graphql schema/resolvers/tests it should be named just `creator`

xoldyckk:
i also noticed many database that must not be nullable but they are, recheck the schema for all such instances, fix their mongoose schema, typescript types and graphql schema types

palisadoes:
Not including `updatedBy` is OK given what we now know

palisadoes:
@LakshyaSatpal Is this ready to be merged? Have you satisfied the needs of all the reviewers?

LakshyaSatpal:
> @LakshyaSatpal Is this ready to be merged? Have you satisfied the needs of all the reviewers?

Yes @palisadoes , I've done the requested changes.

palisadoes:
1. @EshaanAgg
2. @sumitra19jha
3. @xoldyckk

 
Please give this a final review

src/typeDefs/types.ts
```diff
@@ -81,15 +87,20 @@ export const types = gql`
...
     amount: Float!
+    createdAt: DateTime
+    updatedAt: DateTime
   }
   type Advertisement {
-    _id: ID
+    _id: ID!
     name: String!
-    orgId: ID
+    orgId: ID!
```
xoldyckk, line 96: regarding `orgId` field take a look at this https://github.com/PalisadoesFoundation/talawa-api/pull/1585#discussion_r1445939073

src/typeDefs/types.ts
```diff
@@ -35,13 +37,14 @@ export const types = gql`
...
   type Comment {
-    _id: ID
+    _id: ID!
     text: String!
-    createdAt: DateTime
-    creator: User!
     post: Post!
     likedBy: [User]
     likeCount: Int
+    creator: User!
```
xoldyckk, line 45: make the creator field nullable

this isn't limited to this field, by making the related object fields non nullable you make your response object susceptible to error propagation that turns the whole graphql response into null instead of just turning the field that errored out to be null

src/typeDefs/types.ts
```diff
@@ -81,15 +87,20 @@ export const types = gql`
...
+    _id: ID!
     name: String!
-    orgId: ID
+    orgId: ID!
     link: String!
     type: String!
     startDate: Date!
     endDate: Date!
+    createdAt: DateTime
+    creator: User!
```
xoldyckk, line 102: make the creator nullable

src/typeDefs/types.ts
```diff
@@ -25,6 +25,8 @@ export const types = gql`
     user: User!
     event: Event!
     feedbackSubmitted: Boolean!
+    createdAt: DateTime
+    updatedAt: DateTime
```
xoldyckk, line 29: make createdAt and updatedAt non nullable in `type` fields, they will never be null in database, if there has never been an update to a database object after it's creation, its createdAt and updatedAt fields will be equal

src/typeDefs/inputs.ts
```diff
@@ -147,7 +147,7 @@ export const inputs = gql`
     location: String
     attendees: String
     isPublic: Boolean!
-    visibleInSearch: Boolean!
+    visibleInSearch: Boolean
```
xoldyckk, line 150: make this non nullable

xoldyckk:
at this point i'm instructing to correct problems that are unrelated to this PR and are inherent to the codebase 😅

src/typeDefs/types.ts
LakshyaSatpal: I assume, I will also have to remove `required: true` from mongoose schema, to align database schema and graphql schema

LakshyaSatpal: got it @xoldyckk , thank you for the good explanation.
I kept creatorId as `required: true` in the database schema

LakshyaSatpal:
@palisadoes I see some unusual merge conflicts here. The `Task` and `EventProject` model files are deleted. Is this some issue?

palisadoes:
> @palisadoes I see some unusual merge conflicts here. The `Task` and `EventProject` model are deleted. Is this some issue?

We recently removed those from the repo. We'll be implementing `ActionItems` as a replacement.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1648

LakshyaSatpal:
Oh, okay. Will resolve the conflicts

palisadoes:
The counting file changed test is failing because it is a large PR. It was added to alert reviewers that something could be wrong.

This is ready for review.

palisadoes:
> at this point i'm instructing to correct problems that are unrelated to this PR and are inherent to the codebase 😅, but they exist in the code that is concerned within this PR
> 
> @palisadoes should I just ignore them?

Please create an issue to fix them outside of this PR. You have mentioned this a few times in the past. It's best to have another PR decicated to this type of code hygiene

palisadoes:
@EshaanAgg
@sumitra19jha
@xoldyckk

Please take another look

src/typeDefs/types.ts
```diff
@@ -202,14 +223,15 @@ export const types = gql`
     description: String!
     location: String
     isPublic: Boolean!
-    creator: User!
+    creator: User
+    createdAt: DateTime!
+    updatedAt: DateTime!
     members: [User]
-    admins(adminId: ID): [User]
+    admins(adminId: ID): [User!]!
```
xoldyckk, line 230: `[User!]`

src/typeDefs/types.ts
```diff
@@ -202,14 +223,15 @@ export const types = gql`
...
     membershipRequests: [MembershipRequest]
     blockedUsers: [User]
     visibleInSearch: Boolean!
-    apiUrl: URL!
-    createdAt: DateTime
+    apiUrl: URL
```
xoldyckk, line 234: make this `apiUrl: URL!`

`URL` is a graphql scalar, it's just a normal string that is validated within the schema to be some kind of url

src/typeDefs/types.ts
```diff
@@ -233,9 +255,9 @@ export const types = gql`
     name: String!
     description: String!
     isPublic: Boolean!
-    creator: User!
-    visibleInSearch: Boolean!
-    apiUrl: URL!
+    creator: User
+    visibleInSearch: Boolean
+    apiUrl: URL
```
xoldyckk, line 260: `apiUrl: URL!`

src/typeDefs/types.ts
```diff
@@ -233,9 +255,9 @@ export const types = gql`
     name: String!
     description: String!
     isPublic: Boolean!
-    creator: User!
-    visibleInSearch: Boolean!
-    apiUrl: URL!
+    creator: User
+    visibleInSearch: Boolean
```
xoldyckk, line 259: `visibleInSearch: Boolean!`

src/typeDefs/types.ts
```diff
@@ -268,24 +290,25 @@ export const types = gql`
     pluginName: String!
     pluginCreatedBy: String!
     pluginDesc: String!
-    uninstalledOrgs: [ID!]!
+    uninstalledOrgs: [ID]
```
xoldyckk, line 293: `[ID!]`

src/typeDefs/types.ts
```diff
@@ -393,7 +417,7 @@ export const types = gql`
   type UserTag {
     _id: ID!
     name: String!
-    organization: Organization
+    organization: Organization!
```
xoldyckk, line 420: `organization: Organization`

src/typeDefs/types.ts
```diff
@@ -369,7 +392,8 @@ export const types = gql`
       organizationId: ID
     ): UserTagsConnection
     tokenVersion: Int!
-    userType: String
+    updatedAt: DateTime!
+    userType: String!
```
xoldyckk, line 396: make `UserType` graphql enum for this with SCREAMING_SNAKE_CASE named instances

src/typeDefs/types.ts
```diff
@@ -102,7 +113,7 @@ export const types = gql`
     title: String!
     description: String!
     startDate: Date!
-    endDate: Date!
+    endDate: Date
```
xoldyckk, line 116: is this expected to have a nullable endDate?

LakshyaSatpal: I guess, yes, an event may have a null as endDate. @palisadoes?

src/typeDefs/types.ts
```diff
@@ -113,12 +124,14 @@ export const types = gql`
     location: String
     latitude: Latitude
     longitude: Longitude
-    organization: Organization
-    creator: User!
-    attendees: [User!]!
+    organization: Organization!
```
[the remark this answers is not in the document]
palisadoes: There is another open PR for recurring meetings with no end date. It ends then by default in the year 2100.

Based on that we'll need a mandatory end date and let the recurring event logic do the rest.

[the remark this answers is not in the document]
LakshyaSatpal: okay, making endDate non-nullable

xoldyckk (approved):

src/typeDefs/types.ts
```diff
@@ -201,15 +222,17 @@ export const types = gql`
     name: String!
     description: String!
     location: String
-    creator: User!
+    isPublic: Boolean!
```
xoldyckk, line 225: same as above

schema.graphql
```diff
@@ -542,19 +563,21 @@ input OTPInput {
...
   apiUrl: URL!
   blockedUsers: [User]
-  createdAt: DateTime
-  creator: User!
+  createdAt: DateTime!
+  creator: User
   customFields: [OrganizationCustomField!]!
   description: String!
   image: String
+  isPublic: Boolean!
```
xoldyckk, line 574: remove isPublic

palisadoes:
@xoldyckk Please take another look. This is almost complete

xoldyckk (approved):

palisadoes (approved):

Issue #1517, opened 2023-12-20
Upgrade package `uuid`


palisadoes:
**Describe the bug**
When installing the application we get this error on the console

```
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

The package needs to be upgraded

**To Reproduce**
- Run the `npm install` command

**Expected behavior**
1. The error message goes away after the upgrade. No error message must be suppressed
    1. Other dependencies on this package need to be upgraded as recommended by their package maintainers
    2. Deprecated packages depending on this package will need to be replaced with currently maintained equivalents.
    3. Package updates must only be applied to the `package.json` file
1. You may need to update the code base to reflect the updated and/or replaced packages
1. All PR tests pass
1. No functionality is affected by the change
1. You must verify whether the package is still used at all and if not, remove it from any configuration files

**Actual behavior**
- See above

**Screenshots**
- N/A

**Additional details**
- N/A

**Potential internship candidates**
Please read this if you are planning to apply for a Palisadoes Foundation internship https://github.com/PalisadoesFoundation/talawa/issues/359

akhilender-bongirwar:
I would like to work on this issue. Please assign me : ) 
// @noman2002

Veer0x1:
I want to work on this.

techmannih:
@palisadoes ,
I want to work on this issue please assign me

dhruvgupta0503:
hi, i want to work on this issue

AdityaRaimec22:
@palisadoes can I get assigned to this issue.

akhilender-bongirwar:
@Cioppolo14, I am unassigning myself as I don't want to go against our community policy, which states assigning no more than two issues to each contributor across all repositories. Thank you

Veer0x1:
@Cioppolo14 can I work on it.?

palisadoes:
No activity. Removing assignment

shahzamaahmad:
I want to work on this.

NayOoLwin5:
Can you pls assign it to me? I would like to work it

Harsh2563:
I would like to work on this. please assign me this.

palisadoes:
Removing issue assignment. No activity. No PR.

Shubh152:
Please assign this issue to me. @Cioppolo14

[...]

PR #1579, opened 2023-12-27
Upgraded package UUID

dhruvgupta0503:
After running npm install no such error is their

bugfix- earlier the package uuid was not on the latest version , so i had updated uuid with the latest version along with some packages that were dependent on the uuid , for prope functioning 

**Issue Number:**#1517

Fixes #

**Snapshots/Videos:**

**Summary**

**Other information**

YES


package.json
```diff
@@ -82,9 +83,10 @@
     "morgan": "^1.10.0",
     "nanoid": "^3.3.4",
     "nodemailer": "^6.7.8",
-    "pm2": "^5.2.0",
+    "pm2": "^5.3.0",
+    "request": "^2.88.2",
```
xoldyckk, line 87: what's the use of request package?

dhruvgupta0503: provides central logging and help Node.js application running continuously
It also supports load balancing, while updating UUID this package was dependent so in order to fix this bug, this package needs to be upgraded to its latest version

dhruvgupta0503: it helps in making http request easier and while upgrading uuid this package was required to be upgraded 
since it has been deprecated, now axios are commonly used for the same 
but uuid was dependent on the same 
so thats why it was needed to upgrade it

package.json
```diff
@@ -82,9 +83,10 @@
...
     "shortid": "^2.2.16",
-    "uuid": "^9.0.0",
+    "uuid": "^9.0.1",
```
sumitra19jha, line 89: Since this PR was related to UUID version change, Can you show the error or other message which made you change version of pm2, image-hash, and add request

sumitra19jha (commented):
Please provide screenshot of error or some other message which made you add request and change version of image-hash and pm2

package.json
dhruvgupta0503: yess sure

dhruvgupta0503: Screenshot 2024-01-06 003505

I had upgraded all those packages which are mentioned above
pls check and suggest me if they wants to be upgraded or not

palisadoes:
@sumitra19jha @xoldyckk Please one last look

package.json
xoldyckk: why would uuid depend on request package? uuid has nothing to do with http requests

xoldyckk:
@palisadoes uuid v3 is used by these packages

Used by pm2, an archived repository not maintained anymore:-
https://github.com/census-instrumentation/opencensus-node

Some bespoke package with 1300 weekly downloads on npm:-
https://github.com/danm/image-hash

this is a job for maintainers of the packages, if they can't use up to date packages in their libraries, the libraries shouldn't be used

palisadoes:
@xoldyckk
- I agree that we should not be using deprecated or dormat packages in our repo.

@dhruvgupta0503 
1. How can we migrate away from these packages in our code base?
1. This should be the true objective of this PR now that this has been discovered

dhruvgupta0503:
@palisadoes 
As far as I know we can update package versions, that's what I do in the pr and we can find the alternative of those packages which are using similar API to minimize the code change. 
I am looking for the packages which serve the same purpose.

palisadoes:
We need the alternative solution implemented as part of this PR

dhruvgupta0503:
for this we can change the dependencies or we can write that piece of code again 
If you allow me we can write that piece of code again which are using those dependencies

palisadoes:
1. First, we have to close this PR as it's against the `main` branch and not `develop`
2. With the new PR we need a solution where only the `package.json` file is adjusted. If we cannot implement a solution using this, then the issue and subsequent PR will need to be closed.

Please reassess and open an new PR

PR #1603, opened 2023-12-30
Feature: database transaction logging 

### LINKED ISSUE #1480: Feature Request: Create Transaction Log Collection for Database
We don't track when a record is created or updated in the database.

Setting up a new collection dedicated to logging all transactions or changes made within the database. This collection will serve as a detailed history of all data interactions, providing an extensive audit trail.

LakshyaSatpal:
Feature

**Issue Number:**

Fixes #1480

Yes

**Snapshots/Videos:**

**Summary**
Major Changes that this PR introduces:
- a new environment variable `TRANSACTION_LOG_PATH` which the user can set using `setup.ts` or in `.env` manually, if not, default will be `talawa-api/logs/transaction.log`
- changes in `.gitignore`
- a `storeTransaction` utility that stores the transaction in the log file provided. Logs will be written on a line-by-line for every create, update or delete on the database. Example TransactionLog: `timestamp=<example-timestamp> ,createdBy=123, type=UPDATE, model=User, message=User:673512 updated token`
- unit tests for every mutation, that checks if correct logs are added to the log file
- `getTransactionLogs` resolver that parses the log file and returns a list of TransactionLog objects, defaulting to the last 24 hours in descending order of time (latest first old last)

**Other information**
This branch may contain some changes based on #1542. I'd recommend to merge that first, and then this one.


sumitra19jha:
@palisadoes I think its not a good idea to store logs this way. We can create a log folder and save all log there (including Transaction log). Use Cloudwatch or push the log files to s3 on timely basis. One way of implementation is to use morgan.


Example
`
const rTracer = require("cls-rtracer");
const { config } = require("../configs");
const morgan = require("morgan");

const logger = require("pino")({
  level: config.env === "TEST" ? "error" : "info",
  mixin() {
    return { reqId: rTracer.id() };
  },
  transport: {
    target: "pino-pretty",
    options: {
      translateTime: "yyyy-mm-dd HH:MM:ss",
      ignore: "pid,hostname,reqId",
      messageFormat: "[{reqId}]: {msg}",
      colorize: false,
      destination: "./logs.log",
    },
  },
  serializers: {
    req(request) {
      let body = undefined;
      if (request.body && request.headers["content-type"] == "application/json")
        body = request.body;
      return {
        method: request.method,
        url: request.url,
        body,
        ip: request.ip,
        ips: request.ips,
      };
    },
  },
});

const rTracerPlugin = rTracer.fastifyPlugin;

morgan.token("id", (req) => rTracer.id());
morgan.token("response-size", (req, res) => res.get("Content-Length"));
morgan.token("processing-time", (req) => {
  return Date.now() - req.startTime;
});

const morganLogFormat = `
:date[clf] :id ${process.pid} :remote-addr :response-size :response-time ms :processing-time ms :status :url :method :url :http-version
`;

const requestLoggerMiddleware = morgan(morganLogFormat);

module.exports = { logger, rTracerPlugin, requestLoggerMiddleware };

`

sumitra19jha (commented):
I don't think we should proceed with this way of implementation where transaction log is in same db as other data. It will lead to issue in future.

palisadoes:
@LakshyaSatpal Please propose a solution where we use the filesystem instead based on this suggestion.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1603#pullrequestreview-1802895013

palisadoes (changes requested):
@LakshyaSatpal Please propose a solution where we use the filesystem for log output instead based on this suggestion.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1603#pullrequestreview-1802895013

LakshyaSatpal:
Okay. @palisadoes 
one question: Do we still need a resolver like ` getTransactionLogs` that gets the logs from the file system?

palisadoes:
> Okay. @palisadoes one question: Do we still need a resolver like ` getTransactionLogs` that gets the logs from the file system?

1. Yes, that would be useful, defaulting to the most recent time frame.
2. The log location would need to be defined in the `.env` file and `setup.ts`
3. The logs should be a single line per entry when written to disk with standardized formatting with a date and time to the nearest millisecond based on the time of generating the log, not the time that it is written to disk. The single line format makes DevOps basic grepping for matches easier to interpret

setup.ts
```diff
@@ -45,6 +47,28 @@ async function accessAndRefreshTokens(
   }
 }
 
+function transactionLogPath(logPath: string | null): void {
+  const config = dotenv.parse(fs.readFileSync(".env"));
+  config.TRANSACTION_LOG_PATH =
+    logPath ?? path.resolve(__dirname, "logs/transaction.log");
+  fs.writeFileSync(".env", "");
+  for (const key in config) {
+    fs.appendFileSync(".env", `${key}=${config[key]}\n`);
```
palisadoes, line 56: Verify that the file can be written to by the setup user.

setup.ts
```diff
@@ -45,6 +47,28 @@ async function accessAndRefreshTokens(
   }
 }
 
+function transactionLogPath(logPath: string | null): void {
+  const config = dotenv.parse(fs.readFileSync(".env"));
+  config.TRANSACTION_LOG_PATH =
+    logPath ?? path.resolve(__dirname, "logs/transaction.log");
```
palisadoes, line 53: If it's an absolute path, then `logs` must be removed. You are adding to the path without verifying or creating the `logs` directory.

palisadoes (changes requested):
See comments

src/utilities/storeTransaction.ts
```diff
@@ -0,0 +1,30 @@
...
+export const storeTransaction = (
+  createdBy: string | Types.ObjectId,
+  type: string,
+  modelName: string,
+  message: string
+): void => {
+  const logPath = TRANSACTION_LOG_PATH as string;
+
+  const timestamp = new Date().toISOString();
+  const logMessage = `timestamp=${timestamp}, createdBy=${createdBy}, type=${type}, model=${modelName}, message=${message}\n`;
```
palisadoes, line 24: 1. What does the paid version of mongdb log? We should be logging at least that data. 
2. We only have the fact that someone made an update. We don't have any context related to the values updated or deleted. This means a forensic audit is not possible in the case of fraudulent activity. This is insufficient

palisadoes (changes requested):
See comments

src/utilities/storeTransaction.ts
LakshyaSatpal: mongodb logs is actually much detailed, it logs in json format, and example log is something like
`
{
  "timestamp": "2024-01-06T12:34:56.789Z",
  "operation": "update",
  "database": "marketingDB",
  "collection": "advertisements",
  "user": {
    "username": "adminUser",
    "roles": ["dbAdmin"]
  },
  "client": {
    "ip": "192.168.1.100",
    "application": "MongoDB Compass"
  },
  "operationDetails": {
    "query": {"_id": "ad123"},
    "update": {"$set": {"image": "new_image_url.jpg"}},
    "documentId": "ad123"
  },
  "outcome": "success"
}
`
2. I think, for our case, we should atleast add the `update` field, to know what was set/push/pulled from the document. Maybe the client ip too, if possible. What do you think?

sumitra19jha:
@LakshyaSatpal Try to put yourself in the shoe of other programmer. Log codes are almost used at multiple places and program use them to log
- info
- warn
- error

You can add more type based on needs. 

Then there is a message. When you add a log, you type something like
`logger.info("DB Entry added", transaction)`

The more parameter you add, the more you are motivating others to not use. So try to be as minimalist as possible.
Reading that I think can help: https://ritikchourasiya.medium.com/ultimate-guide-about-logging-in-nodejs-and-its-implementation-db593d8ddb54

LakshyaSatpal:
got your point, will refactor this PR, and use winston for logging.

LakshyaSatpal:
@sumitra19jha The problem that I'm facing is that every mutation is different, so even if try to do
`logger.info("User updated", transaction)`. The `transaction` object here needs to manually made based on the mutation, which I think is a burden for other developers who are working on mutations.

One solution is to do this with mongoose middlewares, along with winston. For every, model, I will define a middleware that will automatically log the information without the burden of defining the log information in the mutation. I checked that I go get information that we require in the middlewares my mongoose. 
But, I'm not sure if it's a good practice or it will add more performance burden to database operations?

palisadoes:
> @LakshyaSatpal Try to put yourself in the shoe of other programmer. Log codes are almost used at multiple places and program use them very often. Generally they are classified as:
> 
>     * info
> 
>     * warn
> 
>     * error
> 
> 
> You can add more type based on needs then there is a message. When you add a log, you type something which is as simple as: `logger.info("DB Entry added", transaction)`
> 
> The more parameter you add, the more you are motivating others to not use. So try to be as minimalist as possible. Reading that I think can help: https://ritikchourasiya.medium.com/ultimate-guide-about-logging-in-nodejs-and-its-implementation-db593d8ddb54

@sumitra19jha 

1. We're going to need a very verbose debug level of logging for the rare occasions where developers or users will need very detailed visibility of what is happening. 
2. I agree that in most cases a lower degree of information is desirable, but we'll need this debug option as well.

LakshyaSatpal:
> @sumitra19jha The problem that I'm facing is that every mutation is different, so even if try to do `logger.info("User updated", transaction)`. The `transaction` object here needs to manually made based on the mutation, which I think is a burden for other developers who are working on mutations.
> 
> One solution is to do this with mongoose middlewares, along with winston. For every, model, I will define a middleware that will automatically log the information without the burden of using logger in every mutation. I checked that I do get most of the the logging info that we require in the middlewares by mongoose. But, I'm not sure about 2 things:
> 
> 1. if it's a good practice or it will add more performance burden to database operations?
> 2. If we need the userId of updator in middleware we may need to add a `updatedBy` field to the schema, since I don't get the `context` there. @xoldyckk will need your input on this.

@sumitra19jha @xoldyckk @palisadoes should I be implementing this with mongooose middlewares and winston?

palisadoes:

palisadoes:
1. Please fix the failing `Testing Application` test
2. The counting files test will fail because this is a large PR. It is there to alert reviewers that something could be wrong

setup.ts
LakshyaSatpal: Done

setup.ts
LakshyaSatpal: Done.

LakshyaSatpal:
@palisadoes Application tests are now passing.

palisadoes:
- @xoldyckk
- @sumitra19jha
- @JamaicanFriedChicken

Please review, this is ready

@LakshyaSatpal this may not be reviewed until the weekend due to work commitments

palisadoes:

palisadoes:
A quick reminder. This is a big PR. It will assist greatly in troubleshooting and doing transaction audits.

- @xoldyckk
- @sumitra19jha
- @JamaicanFriedChicken
- @DMills27 

Please review, this is ready

palisadoes:
Also, this PR is blocking the merging of  this code linting PR that checks all updated files. Please review as a priority.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1677

.env.sample
```diff
@@ -49,6 +49,8 @@ SMTP_USERNAME=
 SMTP_PORT=
 SMTP_SSL_TLS=
 
+# Absolute path from storing transaction logs in file system
+TRANSACTION_LOG_PATH=
```
EshaanAgg, line 53: Add a default `log` path here.

schema.graphql
```diff
@@ -861,6 +862,14 @@ input ToggleUserTagAssignInput {
...
 }
 
+type TransactionLog {
+  createdBy: String!
+  message: String!
+  model: String!
+  timestamp: DateTime!
+  type: String!
+}
+
```
EshaanAgg, line 872: Why are we adding `TransactionLog` type to GraphQL schema? Logging is an operation specific to the backend and shouldn't be exposed to clients. Thus, there is no point adding in the schema.

setup.ts
```diff
@@ -46,6 +48,68 @@ async function accessAndRefreshTokens(
   }
 }
 
+function transactionLogPath(logPath: string | null): void {
+  const config = dotenv.parse(fs.readFileSync(".env"));
+  if (!logPath) {
+    // Check if the logs/transaction.log file exists, if not, create it
+    const defaultLogPath = path.resolve(__dirname, "logs");
+    const defaultLogFile = path.join(defaultLogPath, "transaction.log");
```
EshaanAgg, line 56: Notify the developer here that such a directory has been created here for logging.

src/constants.ts
```diff
@@ -412,6 +412,12 @@ export const CUSTOM_FIELD_TYPE_MISSING = {
   PARAM: "customField.isMissing",
 };
 
+export const TRANSACTION_LOG_TYPES = {
+  CREATE: "CREATE",
+  UPDATE: "UPDATE",
+  DELETE: "DELETE",
+};
+
```
EshaanAgg, line 420: Use an Enum and not an object.

src/resolvers/Mutation/acceptAdmin.ts
```diff
@@ -54,5 +55,11 @@ export const acceptAdmin: MutationResolvers["acceptAdmin"] = async (
     }
   );
 
+  await storeTransaction(
+    context.userId,
+    TRANSACTION_LOG_TYPES.UPDATE,
+    "User",
+    `User:${args.id} accepted as admin`
+  );
```
EshaanAgg, line 63: This is not a good approach at all. 
- Logging is backend-specific. If we are awaiting the logging operation in queries, it increases the time our sever takes to respond to each request. Logging should be asynchronous. 
- Can you research a better way of doing this? Adding a log to each file for each transaction is a  lot of work, and difficult to maintain in the upcoming PRs. There should be some middleware or existing package that we can leverage to log all the database changes. [Reference](https://www.mongodb.com/docs/manual/reference/log-messages/)
- These user messages are helpful in debugging! This change is in the right direction, but Im really skeptical about the implementation.
- In the next commit, please make changes according to your new approach in only a couple of files to they are easy to review. Once you get your method approved, you can make the changes to all the files. This would save a lot of effort on you part as well.

src/resolvers/Query/getTransactionLogs.ts
EshaanAgg, line 1: No need to expose the transactional logs to the user. You can add an endpoint to the backend app if you want where the user can see the logs for quick debugging, but the same shouldn't be accessible to clients. It may lead to breach of privacy and data.

EshaanAgg (changes requested):

LakshyaSatpal:
> @sumitra19jha The problem that I'm facing is that every mutation is different, so even if try to do `logger.info("User updated", transaction)`. The `transaction` object here needs to manually made based on the mutation, which I think is a burden for other developers who are working on mutations.
> 
> One solution is to do this with mongoose middlewares, along with winston. For every, model, I will define a middleware that will automatically log the information without the burden of using logger in every mutation. I checked that I do get most of the the logging info that we require in the middlewares by mongoose. But, I'm not sure about 2 things:
> 
> 1. if it's a good practice or it will add more performance burden to database operations?
> 2. If we need the userId of updator in middleware we may need to add a `updatedBy` field to the schema, since I don't get the `context` there. @xoldyckk will need your input on this.
https://github.com/PalisadoesFoundation/talawa-api/pull/1603#issuecomment-1880165751

@EshaanAgg

LakshyaSatpal:
@EshaanAgg 
defined a middleware creation function in `src/libraries/dbLogger.ts`, and used it in inside `src/models/User.ts`. 
If this approach looks good, we can move forward and put the same middleware in all models.

EshaanAgg:
> , along with winston. For every, model, I will define a middleware that will automatically log the information without the burden of using logger in every mutation. I checked that I do get most of the the logging info that we require in the middlewares by mongoose. But, I'm not sure about 2 things:
> 
> 1. if it's a good practice or it will add more performance burden to database operations?
> 2. If we need the userId of updator in middleware we may need

Using middleware is a much better approach IMO. Let's start with only one model, let's say User. Try to develop an asynchronous logger that makes use of Mongoose middleware to do the logging of database transactions automatically. Also add an example that if a develop manually calls `log(x,y,z)` in a resolver, then the same is added to the file. 

Having the code for just one resolver with us would make for a nice POC, and we can decide a further approach. As for your queries, 
1. Sure, logging is definitely a zero-cost operation. That's why we should keep it async as much as possible and in controlled amounts. Allowing users to opt into logging is also a great functionality, so the same is only done in dev environments and not production. 
2. Not really. We can use the middleware provided by Mongoose to log the exact change that was made, and use the express/GraphQL middleware to log the `context.userId` to log the user that made the given request. This approach would allow the separation of concerns, would be more scalable, and allow anybody reading the logs to piece the operations together. 
3. If the logging in async, then make sure the timestamp of changes and requests in also added in the data as the operation of async requests may be different than that they were called in. If we are using synchronous logging, then the same would be moot. 

Again I'd advise, give a day to research all the possible options and only code a small POC, and get them approved before making changes to all codebase. Makes a better experience for both of us! Kudos and thanks for your efforts!

LakshyaSatpal:
@EshaanAgg 
Could you please review the `src/libraries/dbLogger.ts` file and User model file?
Instead of using `fs` module, now I've used the `winston` library to make sure logging is asynchronous

EshaanAgg:
@LakshyaSatpal The made changes looks open. I would like you to do the following:
- Close this PR
- Create a new PR in which you add this `winston` + file based logging option for the users along with the scripts that you created for the same.
- Add middleware for the User mongo object and use the same in any 5-6 user resolvers.
- Attach screenshots of the logs and how they appear in the file when those resolvers are called. 

I feel that should be a sizeable PR that would be easy to review and point out any minor changes. Once the same is finalized, we can quickly implement all the other logic.

LakshyaSatpal:
Okay, Got it @EshaanAgg

LakshyaSatpal:
Closing this PR and making clean PR for the new approach.

PR #1658, opened 2024-01-08
Create weekly event recurring instance

Community-Programmer:
**What kind of change does this PR introduce?**
Fixed the bug on creating recurring event only single event is created

**Issue Number:**

Fixes https://github.com/PalisadoesFoundation/talawa-api/issues/1583

**Did you add tests for your changes?**
yes

**Snapshots/Videos:**

https://github.com/PalisadoesFoundation/talawa-api/assets/76515568/2d723ff1-ae9d-4bfa-a6a7-a7a0597d4c87

**If relevant, did you update the documentation?**

No

**Summary**

- [ ]  Create a weekly recurring instance

**Implementation Details:**

**Validation:**

No other functionalities is affected by changes made in the createEvent  mutation
All test files passed
image

**Does this PR introduce a breaking change?**

Yes

**Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-api/blob/master/CONTRIBUTING.md)?**

Yes

PR #1712, opened 2024-01-17
Create Daily event recurring instance

### LINKED ISSUE #1583: Fix Recurring Events
---
### Note

This is a very important feature to implement. 
1. We need someone with either demonstrable:
    1. experience with GraphQL
    2. activity in their GitHub account with a focus on TypeScript
1. This is not an issue for a first time attempt at the technology. 
    1. The first PR needs to be created within a week
    1. All PRs must be completed by the end of February
1. Read through the comments in the closed PR about the progress and why the approach was selected. There are further details below.

In summary:
1. This is not for a newbie
1. This is not for someone with limited time over the next month
---

**Describe the bug**
The creation of a recurring event only creates a single event in the API clients.

**To Reproduce**

1. Create a recurring event
3. Only a single instance is visible

**Expected behavior**

1. You'll need to update or create new mutations that include start and end date for the event recurrence, not just the event itself
1. There will need to be a query created or updated that would return all the calendar entries for a particular time frame including the recurring ones.
1. We will need the ability to edit any recurrence of the event at any time that the event occurs. The edits would need to be valid for:
    1. All instances in the series
    2. Just the instance in question
    4. The instance in question and all those in future
1. Administrators will need to be able to track attendance for any type of event. Therefore you will need to think of ways to create permanence for historical recurring events for this purpose
1. We don't want the creation of an infinite number of event calendar entries if a recurring event has no end date
1. Other functionality must not be affected
1. You must write tests to validate your work

**Actual behavior**
- See above

**Screenshots**
- N/A

**Additional details**
1. Work with @xoldyckk on evaluating your solution here and in the subsequent PR
1. This is a prerequisite for our menu of project ideas for GSoC 2024
1.  When this issue is resolved we'll need to create another one to support custom recurrences, such as `3rd week of the month`, similar to what is available in Google Calendar
1. We almost completed the work for Weekly recurring when the assignee @Community-Programmer was unable to continue due to exam time constraints. 
    1. The first PR for this issue will need to build on the work in this abandoned PR as it was almost approved for merging. 
        1. https://github.com/PalisadoesFoundation/talawa-api/pull/1738
    1. You will need to submit a plan as described here
         1. https://github.com/PalisadoesFoundation/talawa-api/pull/1738#issuecomment-1910724294
1. You will need to split this issue into multiple PRs, each with custom options provided by the RRULE package
    1. Weekly recurring
    2. Daily recurring
    3. Monthly recurring
    5. Annually recurring

- Please read this if you are planning to apply for a Palisadoes Foundation internship https://github.com/PalisadoesFoundation/talawa/issues/359

Community-Programmer:
Fixed the bug on creating recurring event only single event is created

**Issue Number:**

Fixes #1583 

yes

**Snapshots/Videos:**

https://github.com/PalisadoesFoundation/talawa-api/assets/76515568/688125d5-bec7-423e-82be-361135d5041c

No

**Summary**
Task-2
- [ ] Create Daily event recurring instance

**Implementation Details:**
**Validation:**

No other functionalities is affected by changes made in the createEvent mutation
All test files passed

image

Yes

Yes


palisadoes:
This is a Daily recurring event follow up PR to the Weekly calendar PR merged earlier today.
- https://github.com/PalisadoesFoundation/talawa-api/pull/1658

Please take a look:

1. @xoldyckk
1. @sumitra19jha
1. @DMills27
1. @JamaicanFriedChicken

Community-Programmer:
@palisadoes 
After this, can we have an issue on Talawa Admin? I would like to integrate these features with Talawa Admin
Thank you

palisadoes:
Please review. This is a fix for the daily recurring calendar entry that was never implemented

1. @xoldyckk
1. @sumitra19jha
1. @DMills27
1. @JamaicanFriedChicken
2. @EshaanAgg

palisadoes:
> @palisadoes After this, can we have an issue on Talawa Admin? I would like to integrate these features with Talawa Admin Thank you

Yes, we'll need verification.

src/helpers/eventInstances/daily.ts
```diff
@@ -0,0 +1,50 @@
...
+
+interface InterfaceRecurringEvent extends MutationCreateEventArgs {
+  startDate: Date;
+  creatorId: mongoose.Types.ObjectId;
+  admins: mongoose.Types.ObjectId[];
+  organization: mongoose.Types.ObjectId;
+}
+
+export async function generateEvents(
+  args: Partial<MutationCreateEventArgs>,
```
EshaanAgg, line 19: Why are using `Partial` here? How would you create a valid event if any of the listed `InterfaceRecurrignEvent` args are not passed?

src/helpers/eventInstances/daily.ts
```diff
@@ -0,0 +1,50 @@
...
+  currentUser: InterfaceUser,
+  organization: InterfaceOrganization,
+  session: mongoose.ClientSession
+): Promise<InterfaceEvent[]> {
+  const recurringEvents: InterfaceRecurringEvent[] = [];
+  const { data } = args;
+
+  const startDate = new Date(data?.startDate);
+  const endDate = new Date(data?.endDate);
```
EshaanAgg, line 28: What is the `endDate` is undefined? How many events would then be created? Also if I add a huge end date, let's say `100 years`, then I would be adding `36500` events into the database at one go, that too for one event for just one user. This will completely blow up the database if the organizations start to use this feature leniently. How would you account for these issues?

src/helpers/eventInstances/daily.ts
```diff
@@ -0,0 +1,50 @@
...
+      creatorId: currentUser._id,
+      admins: [currentUser._id],
+      organization: organization._id,
+    };
+
+    recurringEvents.push(createdEvent);
+  });
+
+  //Bulk insertion in database
+  const weeklyEvents = await Event.insertMany(recurringEvents, { session });
```
EshaanAgg, line 47: Why are you calling this `weeklyEvents`?

EshaanAgg (changes requested):
The code is functioning, but it certainly is not scalable. Even if a user enters a date too far into the future by mistake, this will blow up our database with events that we wouldn't be using anytime soon. We need to work on a dynamic strategy that only generates these events as and when required. You need to research how other large calendar based products like `Google Calender`, `Taskade` etc. store and query the same efficiently.

EshaanAgg:
You certainly don't need to add all the functionality in one PR, but you need to first advice a strategy that accounts for all the future and edge use cases. Then we can work on incorporating each functionality in a step by step manner.

Community-Programmer:
@EshaanAgg 

Thank you for the thorough review and insightful suggestions. Your points on scalability and potential database issues were very valid. After conducting further research, I discovered that Google Calendar employs a similar approach by limiting event creation to `100 years` into the future.

Can we follow the same?

By restricting event creation to a reasonable timeframe, we can mitigate the risk of database overload and ensure a more scalable solution.

EshaanAgg:
Can you share the link for the same? 
Also, even if Google does it, I don't feel the same is still feasible for us. They have dedicated SQL databases for the same, while most developers we cater to use the free Mongo Atlas cluster. We need to work on a strategy for the dynamic creation of events based on the pattern of their generation. Read [this discussion](https://stackoverflow.com/questions/5183630/calendar-recurring-repeating-events-best-storage-method) and try to devise a strategy that works for us.

palisadoes:
@EshaanAgg 

1. We recently merged the equivalent for weekly recurring. 1. So we'll need to adjust that functionality to match the outcome of this discussion.

EshaanAgg:
Don't stick to the link I shared. There are many other blog posts about the same. Please try to read as many as you can and come up with a strategy for efficient querying. Please also share all the links that guided your design when sharing, so that even we can vet them once.

Community-Programmer:
> @EshaanAgg
> 
> 1. We recently merged the equivalent for weekly recurring. 1. So we'll need to adjust that functionality to match the outcome of this discussion.



> 1. We recently merged the equivalent for weekly recurring. 1. So we'll need to adjust that functionality to match the outcome of this discussion.

@palisadoes 
I will make the necessary adjustments to the functionality for weekly recurring events to align with the outcome of this discussion in the same pull request.

EshaanAgg:
> @EshaanAgg
> 
> 1. We recently merged the equivalent for weekly recurring. 1. So we'll need to adjust that functionality to match the outcome of this discussion.

I see. Either we should revert it, or we can let it be dormant. When this discussion is completed, we can update it's code too. The ideal solution once achieved should allow for all levels of flexibility in recuraance patterns without much additional configuration.

Community-Programmer:
> Don't stick to the link I shared. There are many other blog posts about the same. Please try to read as many as you can and come up with a strategy for efficient querying. Please also share all the links that guided your design when sharing, so that even we can vet them once.

@EshaanAgg 
Okay, I'll conduct thorough research by reading various blog posts on efficient querying. I'll formulate a strategy based on the insights gathered. When sharing the design, I'll include all the relevant links so that we can collectively review and vet them. I'll keep you updated on my progress.

please have a look at this once
https://stackoverflow.com/questions/50204299/recurring-events-database-model

Community-Programmer:
@EshaanAgg @palisadoes 

We can follow various approaches for managing events:

- We can store cron expressions instead of storing each occurrence or follow a pattern-based recurrence. However, this may lead to a more complex structure and querying process.
-
- The main issue arises when we need to edit any recurrence of the event at any time it occurs, and track attendance for any type of event, etc.

I believe for this scenario, a denormalized structure is better than storing each occurrence separately. One approach could be to reduce constraints, such as allowing the creation of events for only a 10-year period.

what your views on it ?

palisadoes:
Regarding the recurrence limits:

1. The limit cannot be solely date limited. 
2. 100 years of monthly events is very different from 100 years of daily events.
3. There will need to be a sequence count limitation too that would depend on the literature you find.

I'm not a professional coder, so this should be taken into account for my next suggestion. I'd like to hear your feedback. 

1. I think the denormalized approach is simpler to implement, though more of a data tax. 
     1. We link each event in a series to a recurrence ID
     2. When a recurring event is edited you have the option of:
         1.  Updating all event entries with the same ID
         2.  Updating just this event in the series, by giving it a new ID and leaving the rest with the old ID.
         3. There will need to be a way to track historical IDs because:
             1. you may make a change in today's event (New ID #1) and save
             2. you change your mind, and make another change in today's event (New ID#2), but want it reflected in all future events too, but without an ID history it's impossible to do.
                 1. A linked list of events could solve this too

This was the approach recommended here:
- https://stackoverflow.com/questions/50204299/recurring-events-database-model

You could also do a representation of a recurring event as a start date with an instance count tracker to save space. But that would be compute intensive when people scroll through calendars and the API has to generate recurring events in real time while simultaneously presenting non-recurring events.

In the issue discussion thread we thought it would be easier to make all events recurring, even one-time events (0 repetitions), as it would:

1. create standardized logic and;
1. facilitate a single API call.

The editing process would similarly need a historical tracker for the same reasons I mentioned previously.

Community-Programmer:
Absolutely, I agree with your thoughts on the denormalized approach for managing recurring events. The idea of incorporating a sequence count limitation, especially for daily events over a 5-10 year period, seems like a sensible enhancement.I should have applied this approach previously but I didn't know about database constraints.

Treating every event as recurring not only simplifies the logic but also streamlines the codebase. 
Moreover, the denormalized structure indeed ensures efficient updates and makes it feasible to track historical changes effectively. It's a well-rounded approach that addresses both practical considerations and maintainability.

Overall I think it can be a good call!
Need reviews from others also on this approach

Community-Programmer:
@EshaanAgg @palisadoes 

After conducting in-depth analysis and research, I have an approach to use a denormalized structure with minimal impact on the database, particularly for MongoDB, which operates at the document level. This strategy is crucial as it involves managing and tracking edits for each event.

**Approach:**

- Initially, we categorize all events as non-recurring, `ONCE` (0 repetition).

- For events with `WEEKLY, MONTHLY, and YEARLY` recurrence patterns, we set the default endDate as `2124-12-31`, implying a duration of up to `100 years` from the current year. For instance, considering monthly events, there can be a maximum of 5 weekly events in a week and 12 events in a year. Over `100 years`, this results in approximately `12 * 5 * 100`, totaling around `6000` events.

- For `DAILY` events, we allow users to create daily events for up to `10 years` only, generating around `365 * 10`, which amounts to `3650 events`. Additionally, considering that the size of all events will be in kilobytes, this quantity of events is manageable. Users attempting to create daily events beyond the 10-year limit will receive an error message, advising them of the constraint.

By implementing this approach, we effectively limit the number of events stored, ensuring efficient data storage and query performance

**Articles:**

https://scullwm.medium.com/design-and-manage-recurring-events-fb43676e711a

I have also come across another method mentioned in the article that involves the exclusion of a particular event. However, this approach presents a challenge as it prevents us from tracking all events.

https://news.ycombinator.com/item?id=18477975

I would appreciate your input on this matter so that we can move forward and successfully implement this feature within the given time frame.

palisadoes:
@Community-Programmer @EshaanAgg 

The RFP (https://www.ietf.org/rfc/rfc2445.txt) referenced in the Medium article has the basics of what we were talking about in creating a recurring instance tracker.
```
RRULE:FREQ=YEARLY;BYDAY=1SU;BYMONTH=4;UNTIL=19980404T070000Z
```
Recurring events have an RRULE entry inserted in the event and it looks like they use a linked list format too, as I was thinking. With a sequence of multiple recurring rules each with a specified start and end date.

image

The RRULE also allows you specify days of the week etc.

I even found this NPM package. Most of our work may be solved. Just use the schema it recommends.
- https://www.npmjs.com/package/rrule

I vote for the RFP approach. It's what iCal is based on.

palisadoes:
1. It looks like it would simplify things a lot. You don't have to worry about the logic, it's already done. It generates all the dates automatically.
2. The RPF seems to explain how to keep track of successive modifications of recurring instances. When an event is changed, you'll need to modify all the future events previously created by the RRULE.
3. Let us know if this methodology is better and faster to implement than what you had before.

Community-Programmer:
I think it will somehow help us to create recurring events logic and will give us recurrance dates.
main question is that how we are going to handle a single event in recurring series if want to change title, description or anything  related to event?

palisadoes:
In this case, couldn't you replace the existing document with a new one that points back to the original recurring instance?

Community-Programmer:
Do we need to add exclusions for the day that is edited or updated, and should it point to the parent document?

We can certainly do that, but using `rrule `might have a significant impact on the overall project. 

**Here are some considerations:**

- Firstly, we have the eventDashboard for every event, which involves adding projects and attendees.
- Additionally, we have more issues running such as adding volunteer groups and event agendas, making it even more complex.
- Tracking attendance for any type of event will also become difficult.
- If a user edits the event with the option to affect following events, we would need to add exclusions for all those events. This could lead to complex mongoose documents.

Implementing this will take some time, and we will need to make changes to several functions. There will be complex queries, and a majority of functionalities all over the project will be affected and need adjustments accordingly.

palisadoes:
Why can't we just use RRULE to easily generate the dates of the event, then we use your logic to create the events? That way you don't have to worry about the logic.

A recurring event would consist of a series of one or more RRULE objects. 

A series of RRULE objects would contain a recurring event ID that would be back referenced from the event objects.

I'd think if we keep the denormaliszed generation of events like the original design we'd be OK.

Community-Programmer:
Okay, got it. So, we are continuing with the denormalized generation of events, just like we did before. All I have to do is use RRULE to generate the recurring event instances to simplify the logic , right?

palisadoes:
Would this make it easier based on your research?

If not, well have to try something else.

Community-Programmer:
Yeah, in terms of generating events, it is a lot easier. The library you suggested is just awesome for getting recurring dates and will overall reduce complexity and the codebase.
It also has a demo gh-page, which will help to get occurrences, rrule string much faster and in an easier way.
https://jkbrzt.github.io/rrule/

**Only one main question:**
 What to use, a normalized data model or denormalized?

In a normalized data model, we need one more collection named 'rrule,' as you mentioned, and it will refer to each recurring instance. I think we can go with this, and I believe it will not take much time to implement

What are your thoughts on this? According to you, which is more suitable?

palisadoes:
If you're mean that the RRULE collection would be normalized but it would be used to create denormalized events to reduce compute loads when scrolling though calendars, then that makes sense.

Let's see your schema update

Community-Programmer:
> If you're mean that the RRULE collection would be normalized but it would be used to create denormalized events to reduce compute loads when scrolling though calendars, then that makes sense.
> 
> Let's see your schema update

I will look something like this

```javascript
RRULE Object:
{
  RRULE_ID: "abc123",
  Frequency: "Weekly",
  Interval: 1,
  StartDate: "2024-01-01",
  // Other RRULE properties...
}

Event Instances:
{
  EventID: "event001",
  RRULE_ID: "abc123",
  Date: "2024-01-15",
  // Other event details...
}

{
  EventID: "event002",
  RRULE_ID: "abc123",
  Date: "2024-01-22",
  // Other event details...
}
```

Community-Programmer:
Yes exactly, we will be normalizing the RRULE collection to efficiently store recurrence rules. Subsequently, we'll leverage this normalized data to create denormalized events. 

This approach will significantly reduce the computational load on the server, especially during operations like scrolling through calendars. Importantly, it won't increase the load when generating events in real-time. Additionally, the use of denormalized events will enhance the efficiency of editing, particularly for events with the same `RRULE ID`. 

By incorporating these additions and the constraints mentioned in my previous comments, this strategy achieves a balance between data organization, retrieval efficiency, and overall system performance. 
Consequently, there's no need to worry about server load, infinite event generation, or storage concerns.

Lets decide it fast and get it done within the given time frame

palisadoes:
Go ahead

Community-Programmer:
@EshaanAgg 

The approach you proposed for handling recurring events is good, but there are some challenges and issues with this that might act as stoppers. 

Please see the points below and tell what you think:

- Generation of events based on queries might introduce computational overhead which will lead to increased response times for users, especially if there are frequent and complex queries. This could impact system performance and overall user experience.

- Also if events are generated , there could be scenarios where multiple queries or operations are attempting to create or update events concurrently. This concurrency might lead to potential race conditions, where data inconsistencies or conflicts arise due to simultaneous modifications.

- Limiting the pre-generation of events to a maximum of 100 might not suit all scenarios.

- Relying on dynamic generation of events for each query may pose challenges for scenarios where cached data is crucial. In situations where users need to access their events in low-connectivity environments, the absence of pre-generated events could result in incomplete or unavailable data

I think Pre-generating all events could provide a more optimized and responsive solution, ensuring that the required data is readily available when users make queries (since we are caching event also when creating events) , without the need for additional computational effort during each request, though we have some database constraints but we can manage this accordingly by applying constraints for creating event.
Because at end what we need is performance not bad user experience

xoldyckk:
dynamically generating the recurring events seems like the better option

an instance of a recurring event that has been modified by the end user(an exception to the other instances within the same recurring event) should be stored seperately in another table

more info in this article:- https://vertabelo.com/blog/again-and-again-managing-recurring-events-in-a-data-model/

palisadoes:
I agree with @EshaanAgg and @xoldyckk. If we dynamically generate and insert events beyond some global LIMIT, then we reduce DB bloat. It is also unlikely that there will be calendar view overload because the first person who views the event beyond the LIMIT will create the event, thereby relieving the need to do further computation by others viewing the same event.

The extra coding to do this shouldn't be excessive as RRULE will be doing most of the work.

Seeing the updated schema would assist.

palisadoes:
@Community-Programmer, @xoldyckk, @EshaanAgg 

This is my understanding.

1. As before:
   1. Track with an RRULE collection
   1. However, generate up to LIMIT denormalized events
1. New stuff:
   1. Generate additional denormalized events when someone wants to see beyond the previously generated LIMIT number of denormalized events, say some future DATE. Possibly adding LIMIT denormalized events beyond DATE.

palisadoes:
We can't do a pure generation on demand as the events in our series will often be uniquely modified. Possible cases include:

- New invitees
- New volunteers
- New agenda items

With this method, future generated events would copy all the characteristics of the most recently generated event, ie. the event at the LIMIT date.

palisadoes:
@Community-Programmer is this clearer now?

palisadoes:
1. So you are going with a generate events on demand without the LIMIT
2. How will you handle someone changing the date of just one event in the series with your schema?

Community-Programmer:
> 1. So you are going with a generate events on demand without the LIMIT
> 2. How will you handle someone changing the date of just one event in the series with your schema?

@palisadoes 

No, 'not on demand' means that, on event edit, only the event that is edited will be created and stored as a skipped event. This implies that we will not be generating any events until they are edited

Community-Programmer:
Can you please explain the flow? I believe there might be some conflicts in our understanding of the solution. What exactly are you thinking?

palisadoes:
1. Are skipped events the ones that are out of sequence? 
    1. If that's the case then the name you have chosen isn't intuitive. Skipped events sounds like events that should be ignored or not attended. 
    1. Or are they events that should not be in the RRULE series because another out of sequence event has replaced it? If that's the case, how is the new event tracked as part of the series for future updates?
3. Based on your flow, events that are queried will be created on demand if they don't previously exist. Is that correct?

Community-Programmer:
yaa exactly skipped events are events that should not be in the RRULE series as they have been edited or modified so they will be stored as exclusion in the database and refer to the parent_event.

The flow is very simple `only create event that is updated `  rest all will be generated using RRule and same will be returned to  frontend for rendering calendar.

I am talking about below approach 

image

palisadoes:
1. Their use of the `exclusion` terminology makes more sense, but I think `exceptions` is better. It looks like there are two types of `exception` events: 
    1. Events in the series that must not be visible (as in deleted)
    1. Events in the series that have been edited in some way (some other flag)

This means that the RRULE object would need to include an event object to set the parameters to use for the series.

If a user decides to change any parameter in the RRULE string from a point in time onward, how would that work? Would it be like this?

1. Edit the original RRULE
    1. Change the end dateTime
    2. Leave the `exceptions` unchanged or cleared?
1. Create an RRULEnew using:
    1. the point in time event as the seed
    1. an empty `exceptions` list

I just tried this in Google Calendar:
1. Created a recurring daily event (Title: Test Daily)
2. One week later I edited it to become a weekly event (Title: Test Weekly)
3. One month later I created a monthly event (Title: Test Monthly)

Deleting all the weekly events, doesn't delete the daily or monthly. They are handled as completely different recurrences.

This is the link we should reference:
1. https://news.ycombinator.com/item?id=18477975

palisadoes:
@Community-Programmer 

Have you seen this thread? You'll need to ensure that your solution works with the community version of MongoDB. If this is the case, why did your tests pass for the previous PR?
- https://thepalisadoes-dyb6419.slack.com/archives/CSWGQ70UQ/p1705924020861829

Community-Programmer:
@palisadoes 

As discussed earlier, the solution we're trying to implement at starting requires database replication for functioning of transactions. The error you encountered with the standalone instance is due to the absence of replication.

In the previous pull request, the tests passed because we were not starting a session. Transactions were not initiated, which is why the issue didn't arise. 
Rest assured, the approach we're taking now does not rely on transactions, eliminating the need for them in our setup.
Also please revert the previous PR as we are using new approach now

Community-Programmer:
> This is the link we should reference:
> 
> 1. https://news.ycombinator.com/item?id=18477975
> 
> From the discussion:
> 
> 1. Their use of the `exclusion` terminology makes more sense, but I think `exceptions` is better. It looks like there are two types of `exception` events:
>    
>    1. Events in the series that must not be visible (as in deleted)
>    2. Events in the series that have been edited in some way (some other flag)
> 
> This means that the RRULE object would need to include an event object to set the initial parameters to use for the series. This would need to be reflected in your schema.
> 
> If a user decides to change any parameter in the RRULE string from a point in time onward, how would that work? It would seem to be like this, which is validated by my Google Calendar tests
> 
> 1. Edit the original RRULE
>    
>    1. Change the end dateTime
>    2. Leave the `exceptions` unchanged or cleared?
> 2. Create an RRULEnew using:
>    
>    1. the point in time event as the seed
>    2. an empty `exceptions` list
> 
> Google Calendar (Test: 1):
> 
> 1. Created a recurring daily event (Title: Test Daily)
> 2. One week later I edited it to become a weekly event (Title: Test Weekly)
> 3. One month later I edited it again to become a monthly event (Title: Test Monthly)
> 
> Deleting all the weekly events, doesn't delete the daily or monthly. They are handled as completely different recurrences.
> 
> Google Calendar (Test: 2):
> 
> 1. Created a recurring daily event (Title: Test Daily)
> 2. One week later I added somebody to the invitee list
> 3. I deleted the event, all instances, and all entries were deleted.
> 
> Google Calendar (Test: 3):
> 
> 1. Created a recurring daily event (Title: Test Daily)
> 2. One week later I changed the title of the event
> 3. I deleted the event, all instances, and all entries were deleted.
> 
> It appears Google only creates a new recurrence instance when the recurrence definition changes.
> 
> This looks fairly straightforward to implement. Please post the updated schema.

I believe only minimal changes are needed in the schema. Adding just one field for the `'rrule' string` and another for `exception events` should be sufficient to handle it.
And rest RRULE will manage as we have option to exclude a date
image

palisadoes:
> @palisadoes
> 
> As discussed earlier, the solution we're trying to implement at starting requires database replication for functioning of transactions. The error you encountered with the standalone instance is due to the absence of replication.
> 
> In the previous pull request, the tests passed because we were not starting a session. Transactions were not initiated, which is why the issue didn't arise. Rest assured, the approach we're taking now does not rely on transactions, eliminating the need for them in our setup. Also please revert the previous PR as we are using new approach now

The PR cannot be reverted it's too late. So you'll have to update those files in this PR

palisadoes:
OK, let's proceed as we now have a way to minimize the data footprint while optimizing CPU load.

palisadoes:
Adding for cross referencing
- https://github.com/PalisadoesFoundation/talawa-api/pull/1658

Community-Programmer:
> > @palisadoes
> > As discussed earlier, the solution we're trying to implement at starting requires database replication for functioning of transactions. The error you encountered with the standalone instance is due to the absence of replication.
> > In the previous pull request, the tests passed because we were not starting a session. Transactions were not initiated, which is why the issue didn't arise. Rest assured, the approach we're taking now does not rely on transactions, eliminating the need for them in our setup. Also please revert the previous PR as we are using new approach now
> 
> The PR cannot be reverted it's too late. So you'll have to update those files in this PR

OK, I will make all the changes in this pull request only

Community-Programmer:
> OK, let's proceed as we now have a way to minimize the data footprint while optimizing CPU load.

@palisadoes @xoldyckk @EshaanAgg 

OK, so I am proceeding now with my first steps:

- Updating the createEvent schema by adding a field for rruleString.
- Creating a query that will return all the events based on RRULE to the frontend.

Afterwards, I will focus on editing events and adding an exception for the edited and modified event

palisadoes:
@Community-Programmer 

1. We have discussed a lot of logic in this thread which will be useful for other developers. 
2. Please add a detailed description of the workflow in the TypeDoc section of `createEvent` in `src/resolvers/Mutation/createEvent.ts` as part of this PR

Community-Programmer:
Sorry,the force push was an unintended action, that resulted in the automatic closure of the pull request (PR). I understand the inconvenience and disruption this may have caused.

Thank you for your understanding

PR #1883, opened 2024-02-22
added graphql connection utilities

xoldyckk:

feature

**Issue Number:**

-

Yes

**Snapshots/Videos:**

-

-

**Summary**

Currently there's no standard way to implement graphQL connections in talawa-api. Each connection resolver would need to duplicate some logic which can be abstracted away to ease the flow of implementing the connections.

This PR provides some simple utilities to create graphQL connections which cover most of the common use cases. Some special cases might require additional logic which isn't covered in this PR, those special cases will have to be handled for that connection resolver seperately when the time calls for it. There is never a one size fits all solution.

It might be a breaking change if the clients are making use of the the graphQL resolvers and typeDefs that have been modified. Though I don't think that's the case. 

**Other information**

-

Yes


xoldyckk:
Forgot to add `totalCount` field.

xoldyckk:
@palisadoes ready for review

palisadoes:
Tests are failing. Please check

xoldyckk:
done, the file count workflow check is kind of silly in my opinion, I don't want to have to make seperate PRs to just reduce the file count

palisadoes:
We have the file count check to help alert us to when we get incorrect merges. The 20 count is probably too low.

palisadoes (approved):

AVtheking:
@xoldyckk  sir what is the signifigance of parse cursor function I didn't get it ?

PR #2257, opened 2024-04-23
Mongodb setup bug + tests

### LINKED ISSUE #2255: Tests: src/setup/MongoDB.ts
 
 1. Tests need to be written for file `src/setup/MongoDB.ts`
 
image

 2. We will need the API to be refactored for all methods, classes and/or functions found in this file for testing to be correctly executed.
 3. When complete, all all methods, classes and/or functions in the refactored file will need to be tested.
 4. These tests must be placed in a single file with the same path as the tested file, but under the `tests/` with a `.spec.ts` suffix.
 5. You may need to create the appropriate directory structure to do this.
 
 ### IMPORTANT:
 
 * [Create all API Resolver Tests #490](https://github.com/PalisadoesFoundation/talawa-api/issues/490)
 
 ### PR Acceptance Criteria
 * When complete this file must show **100%** coverage when merged into the code base. This will be clearly visible when you submit your PR.
 * [The current code coverage for the file can be found by visting this page](https://app.codecov.io/gh/PalisadoesFoundation/talawa-api?displayType=list). Login using your GitHub credentials.
 * Create your own `codecov.io` to help with testing.
 * The PR you create will show a report for the code coverage for the file you have added. You can also use that as a guide.
 
 ### Other
 Please coordinate **issue assignment** and **PR reviews** with the contributors listed in this issue:
 
 * [Student Internship Programs talawa#359](https://github.com/PalisadoesFoundation/talawa/issues/359)

pranshugupta54:

Bug fix + added tests

**Issue Number:**

Fixes #2255

yes

**Snapshots/Videos:**

image

NA

**Summary**

- Currently it was checking localhost connection even if user had not set the `env` causing no value in the final env.

**Other information**

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -4,20 +4,23 @@ This document provides instructions on how to set up and start a running instanc
 
 # Table of Contents
 
-<!-- toc -->
+<!-- TOC -->
 
+- [Talawa-API Installation](#talawa-api-installation)
+- [Table of Contents](#table-of-contents)
+- [Installation Steps Summary](#installation-steps-summary)
 - [Prerequisites](#prerequisites)
-  - [Install node.js](#install-nodejs)
-  - [Install npm](#install-npm)
-  - [Install TypeScript](#install-typescript)
   - [Install git](#install-git)
   - [Setting up this repository](#setting-up-this-repository)
+  - [Install node.js](#install-nodejs)
+  - [Install TypeScript](#install-typescript)
 - [Installation Using Docker](#installation-using-docker)
-  - [Prerequisites](#prerequisites-1)
+  - [Run the Talawa-API Setup](#run-the-talawa-api-setup)
+  - [Install the Docker Application](#install-the-docker-application)
   - [Docker Compose Setup](#docker-compose-setup)
     - [For Development](#for-development)
     - [For Production](#for-production)
-    - [Congratulations! 🎉 Your Talawa API is now successfully set up and running using Docker!](#congratulations-%F0%9F%8E%89-your-talawa-api-is-now-successfully-set-up-and-running-using-docker)
+  - [Import The Sample Data](#import-the-sample-data)
 - [Installation without Docker](#installation-without-docker)
   - [Install the Required Packages](#install-the-required-packages)
   - [Install MongoDB](#install-mongodb)
@@ -31,24 +34,24 @@ This document provides instructions on how to set up and start a running instanc
     - [The Environment Variables in `.env`](#the-environment-variables-in-env)
     - [Changing the Development Environment in the .env file](#changing-the-development-environment-in-the-env-file)
     - [Generating Token Secrets](#generating-token-secrets)
-      - [Setting up ACCESS_TOKEN_SECRET in the .env file](#setting-up-access_token_secret-in-the-env-file)
+      - [Setting up ACCESS\_TOKEN\_SECRET in the .env file](#setting-up-access_token_secret-in-the-env-file)
         - [Linux](#linux)
         - [Windows](#windows)
-    - [Setting up REFRESH_TOKEN_SECRET in the .env file](#setting-up-refresh_token_secret-in-the-env-file)
+    - [Setting up REFRESH\_TOKEN\_SECRET in the .env file](#setting-up-refresh_token_secret-in-the-env-file)
     - [Configuring MongoDB in the .env file](#configuring-mongodb-in-the-env-file)
-      - [Using the CLI to get the MONGODB_URL Connection String](#using-the-cli-to-get-the-mongodb_url-connection-string)
-      - [Using Microsoft Windows to get the MONGODB_URL Connection String](#using-microsoft-windows-to-get-the-mongodb_url-connection-string)
+      - [Using the CLI to get the MONGODB\_URL Connection String](#using-the-cli-to-get-the-mongodb_url-connection-string)
+      - [Using Microsoft Windows to get the MONGODB\_URL Connection String](#using-microsoft-windows-to-get-the-mongodb_url-connection-string)
     - [Configuring Redis in .env file](#configuring-redis-in-env-file)
       - [For Local Setup (Linux and WSL)](#for-local-setup-linux-and-wsl)
       - [For Remote Setup (Redis Cloud)](#for-remote-setup-redis-cloud)
-    - [Setting up .env LAST_RESORT_SUPERADMIN_EMAIL parameter in the .env file](#setting-up-env-last_resort_superadmin_email-parameter-in-the-env-file)
+    - [Setting up .env LAST\_RESORT\_SUPERADMIN\_EMAIL parameter in the .env file](#setting-up-env-last_resort_superadmin_email-parameter-in-the-env-file)
     - [Configuring Google ReCAPTCHA in the .env file](#configuring-google-recaptcha-in-the-env-file)
-      - [Setting up the RECAPTCHA_SECRET_KEY](#setting-up-the-recaptcha_secret_key)
-      - [Setting up .env MAIL_USERNAME and MAIL_PASSWORD ReCAPTCHA Parameters](#setting-up-env-mail_username-and-mail_password-recaptcha-parameters)
+      - [Setting up the RECAPTCHA\_SECRET\_KEY](#setting-up-the-recaptcha_secret_key)
+      - [Setting up .env MAIL\_USERNAME and MAIL\_PASSWORD ReCAPTCHA Parameters](#setting-up-env-mail_username-and-mail_password-recaptcha-parameters)
     - [Setting up SMTP Email Variables in the .env File](#setting-up-smtp-email-variables-in-the-env-file)
     - [Setting up Logger configurations](#setting-up-logger-configurations)
-      - [Setting up COLORIZE_LOGS in .env file](#setting-up-colorize_logs-in-env-file)
-      - [Setting up LOG_LEVEL in .env file](#setting-up-log_level-in-env-file)
+      - [Setting up COLORIZE\_LOGS in .env file](#setting-up-colorize_logs-in-env-file)
+      - [Setting up LOG\_LEVEL in .env file](#setting-up-log_level-in-env-file)
 - [Importing Sample Database](#importing-sample-database)
   - [Syntax:](#syntax)
   - [Examples:](#examples)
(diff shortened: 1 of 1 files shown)


pranshugupta54:
Oops, made the PR on `main` branch and it automatically added Peter for review 😅

pranshugupta54:
Added localhost as default instead of testing it ourselves 
image

Cioppolo14:
@pranshugupta54 Please fix the failed tests.

src/setup/MongoDB.ts
```diff
@@ -71,7 +64,7 @@ export async function askForMongoDBUrl(): Promise<string> {
       type: "input",
       name: "url",
       message: "Enter your MongoDB URL:",
-      default: process.env.MONGO_DB_URL,
+      default: "mongodb://localhost:27017",
```
palisadoes, line 67: 1. What happens if the parameter is already defined and it's not `localhost` or port `27017`? Will the application break?
2. Can a ternary be used?

palisadoes (changes requested):
See comments

palisadoes:
Thanks,

1. I really don't want setup to break and though this is a test PR, there are changes to the code. 
2. @Anubhav-2003, @meetulr, @adi790uu and @Doraemon012 have all worked on this. I'd appreciate your comments.

Setup is the first DevOps and developer experience with the code. It needs to be perfect.

src/setup/MongoDB.ts
pranshugupta54: If we have already defined then it's going to show us
image

pranshugupta54:
@palisadoes 
The current implementation is wrong. It's currently testing [env MONGO url and localhost url]. If any of them is found to work properly, it moves to next step. But whats wrong in this? If user did not add the url and left the env like `MONGO_DB_URL=` then it's going to test local url and move forward, without even fixing the env.

Steps:
- Make `MONGO_DB_URL` empty in your env like, `MONGO_DB_URL=`
- Run setup with `npm run setup`
- It's going to show that localhost is detected 
- Continue with it
- Now in the last step for import data, it will throw error that can not connect to DB.
- It did not save anything in env for `MONGO_DB_URL`


What I've done now, 
- If there is empty string `MONGO_DB_URL`, it will ask you to enter a **new url** or use the **default localhost url**
- It's going to check if it can connect to that url and then adds it to the env
- Continues to work

- If `MONGO_DB_URL` exists, it will simply ask if you want to change or not (same like other fields)
- If no change then continues
- If change then checks and adds

meetulr:
This is bad, things are messy here, the way the setup for mongoDB is written is very ambiguous and confusing.

The purpose of `setup.ts` is basically to setup our `.env` file (after checking for things like valid connections, etc.) so that the app would run properly after that, right?
So, the way to go is to check the url where a mongoDB instance is running and add it as `MONGO_DB_URL` in our `.env`.

These are the things that need attention in the current implementation:
1) Ambiguous variable naming and comments: The way the functions and variables are named, comments are written, are the direct cause of confusion.
  - The comment for this function [link](https://github.com/PalisadoesFoundation/talawa-api/blob/7bd0b3f5c5fab60d705567a5a06a3b2f57006278/src/setup/MongoDB.ts#L10, says in one line that it checks if an existing mongoDB instance is running, but in the next line it says its checking for an existing  mongoDB connection (which it's not).
  - It is checking if a mongoDB instance is running on a given mongo URL, if so, it sets that as the `MONGO_DB_URL` in our `.env` so that the app can connect to it when we run it, as I mentioned before.
  - Checking for an existing connection implies whether the connection to that mongoDB instance exists or not, which is wrong. The function [link](https://github.com/PalisadoesFoundation/talawa-api/blob/7bd0b3f5c5fab60d705567a5a06a3b2f57006278/src/setup/MongoDB.ts#L41) is not checking whether a connection exists, it's checking whether a connection can be established (which is the thing to do here).

So, here the comments and function named need to be modified.

2) Improper check: [here](https://github.com/PalisadoesFoundation/talawa-api/blob/7bd0b3f5c5fab60d705567a5a06a3b2f57006278/setup.ts#L825), after finding a mongo url in `.env`, if we select to change the url, it checks whether an instance is running on that or other url or not (the variable name [isConnected](https://github.com/PalisadoesFoundation/talawa-api/blob/7bd0b3f5c5fab60d705567a5a06a3b2f57006278/setup.ts#L408) is again wrong here, i.e. implies that there's a mongo connection established. What it actually represents is that whether a url exists where a mongoDB instance is running). On the other hand, if we select not to change the url, it doesn't check whether an instance is running on that url or not, just moves on to the next step in the setup.

src/setup/MongoDB.ts
meetulr: but if another url is set in our env, it won't show that, it'll always show "mongodb://localhost:27017"
the line should be this `default: process.env.MONGO_DB_URL ?? "mongodb://localhost:27017"`

pranshugupta54:
I agree with the bad naming of variables and comments. I'll probably fix the naming and comments in this PR itself today. 

> On the other hand, if we select not to change the url, it doesn't check whether an instance is running on that url or not, just moves on to the next step in the setup.

@meetulr, we are actually not checking this for any step. If variable is found then we skip for every step like Redis etc. 
All your points are valid except the above, if we add checks for found variables then the setup will take too long.

pranshugupta54:
@meetulr, other than naming and comments - do you find any issue with the fixed implementation? Or PR looks better than current situation

meetulr:
> @meetulr, we are actually not checking this for any step. If variable is found then we skip for every step like Redis etc.
All your points are valid except the above, if we add checks for found variables then the setup will take too long.

The purpose is to setup up a running url right, I'm saying that if a url is found in `.env` and the user opts not to change it, it's not checking whether an instance is running on that url or not, which would cause "connection refused" issue when we run the app after "successfully setting the app". If this is done in other places, they need to be fixed too I guess.

src/setup/MongoDB.ts
pranshugupta54: check the screenshot in previous comment, it's showing different. The port is different from default.

default means - when we have Yes/No type, do you see one of them marked automatically and u press enter?

pranshugupta54:
We're not checking if it's already in env, it will directly be seen when we start the server. There's no point checking for previous values in setup. We don't run setup everytime we start the server so it's just for setting up mostly on the first time.

Will fix the 2nd point, Yes!

src/setup/MongoDB.ts
meetulr: the screenshot is for a different check

pranshugupta54: You said it doesn't show from env. That url is from env itself, check the port. It's different port from default

meetulr:
> We're not checking if it's already in env, it will directly be seen when we start the server.

This is confusing:
1) So, if a url is in the `.env` and the user doesn't change it, we're assuming that the developer knows that it's an active url (having an active mongoDB instance).
2) If the user changes it, we're assuming that the developer doesn't know whether it's an active url or not, and we need to check for that?

> There's no point checking for previous values in setup.

Previous values? setup is meant to run first, and check if things are ready to run. It can be run again, and do the same thing.

> We don't run setup every time we start the server so it's just for setting up mostly on the first time.

exactly, setting up is meant to set things up so that the app can run properly.

src/setup/MongoDB.ts
meetulr: what? we're talking about the line that Peter commented on, you've hard coded `mongodb://localhost:27017` there, so when that function runs, it'll always show `mongodb://localhost:27017` for that prompt.

> You said it doesn't show from env. That url is from env itself, check the port. It's different port from default

the screenshot you've provided is just a log, that logs [link](https://github.com/PalisadoesFoundation/talawa-api/blob/7bd0b3f5c5fab60d705567a5a06a3b2f57006278/setup.ts#L814), it will show what's in the `.env`

Anubhav-2003:
> The purpose is to setup up a running url right, I'm saying that if a url is found in `.env` and the user opts not to change it, it's not checking whether an instance is running on that url or not, which would cause that "connection refused" error when we run the app after "successfully setting up the app". If this is done in other places, they need to be fixed too I guess. It won't take too long, the procedure should be the same right? I think things can be refactored a little here.
> 

look what you are saying, makes complete sense that we should check for whether a connection can be established, even if the user opts not to change the URL. It definitely won’t take a long time, so this is a valid point.

But you people must also realise that this setup was written. I’m not sure maybe more than a year ago. In fact when I worked on the set up a few months ago, my job was to refactor it and write tests. And yes, comments and the variable names are definitely ambiguous sometimes. But again, these were written by someone maybe more than a year ago, so we don’t know what was going on in his mind and maybe these were written by multiple people. The repository was quite different back then. So all we can do is if we find any errors, then we should discuss like @pranshugupta54 took initiative, and fix them.

Anubhav-2003:
> > We're not checking if it's already in env, it will directly be seen when we start the server.
> 
> This is confusing:
> 
> 1. So, if a url is in the `.env` and the developer doesn't change it, we're assuming that the developer knows that it's an active url (having an active mongoDB instance).
> 2. If the developer changes it, we're assuming that the developer doesn't know whether it's an active url or not, and we need to check for that?
> 
I don’t think this is confusing, look the purpose of the set up was to ensure that while you are initially setting up the application, all the variables and parameters provided are working properly in order for the application to start up. 
So if we know that a variable is correct and the developer chooses not to change it, then it is not completely unreasonable to not check it again, assuming that it will be correct. Now again, you can obviously add an extra lay of check. As I wrote above, but the present implementation is not wrong and definitely not confusing.

Anubhav-2003:
@pranshugupta54 The PR looks good to me. But yes the part about hardcoding by @meetulr in the prompt seems correct. But that’s a trivial fix.

pranshugupta54:
@Anubhav-2003, the default value is only used when there is no value given and user presses enter to select default.

Just like when you init a new npm folder then it asks for so many of the things, you can either choose the default or add your value.

Anubhav-2003:
> @Anubhav-2003, the default value is only used when there is no value given and user presses enter to select default.
> 
> Just like when you init a new npm folder then it asks for so many of the things, you can either choose the default or add your value.
> 
> Like, Q) What Port do you want? (4000)
> 
> Now, if you press "Enter/Return" - it will use the default 4000 value. If you start typing something and enter 4321 then it will use 4321 and save it.

Ok understood. Thanks👍

meetulr:
> > The purpose is to setup up a running url right, I'm saying that if a url is found in `.env` and the user opts not to change it, it's not checking whether an instance is running on that url or not, which would cause that "connection refused" error when we run the app after "successfully setting up the app". If this is done in other places, they need to be fixed too I guess. It won't take too long, the procedure should be the same right? I think things can be refactored a little here.
> 
> look what you are saying, makes complete sense that we should check for whether a connection can be established, even if the user opts not to change the URL. It definitely won’t take a long time, so this is a valid point.
> 
> But you people must also realise that this setup was written. I’m not sure maybe more than a year ago. In fact when I worked on the set up a few months ago, my job was to refactor it and write tests. And yes, comments and the variable names are definitely ambiguous sometimes. But again, these were written by someone maybe more than a year ago, so we don’t know what was going on in his mind and maybe these were written by multiple people. The repository was quite different back then. So all we can do is if we find any errors, then we should discuss like @pranshugupta54 took initiative, and fix them.

Brother, the purpose of my comments is not to blame anyone😂, but to point out the things that can be / need to be improved😅.

> look the purpose of the set up was to ensure that while you are initially setting up the application, all the variables and parameters provided are working properly in order for the application to start up.

Exactly, thus my insistence on comprehensive checks.

> So if we know that a variable is correct and the developer chooses not to change it, then it is not completely unreasonable to not check it again, assuming that it will be correct.

I'm saying that if there exist a url and user doesn't change it we're not checking if there's running instance there. That contradicts your first point which is what I was saying.

Assuming part is my concern, when user changes the url, we "assume" that we need to check for a running instance, when they don't, we "assume" that it running. "again"? we're checking for a url for a running instance right?

> Now again, you can obviously add an extra lay of check. As I wrote above, but the present implementation is not wrong and definitely not confusing.

Again, you're looking at it from the perspective of defending the code😂 which no one is blaming you for. It is `definitely not confusing` is definitely a strong comment, because it is if you look at it from my perspective.
Your line of reasoning... obviously, I thought of that too, I wanted to be comprehensive in testing that's why I said we could simply add a check for that too.

Ofcourse, It is not a completely ureasonable assumption, but it is that, an "assumption". We could go with the current implementation assuming that. I Just thought it wouldn't be much work to just check it always.

meetulr:
@pranshugupta54 The default is not just used for when there's no url, it would be promped even when there's no instance running on a provided url, for cases where a user forgot to run the instance before starting the setup script, then, the user would be promped with their provided url (if not hard-coded like it is here), they can simply start the instance and continue the setup process. This could be a very common situation right? This is why I'm saying it should be this `default: process.env.MONGO_DB_URL ?? "mongodb://localhost:27017"`, serves both purposes, and I see nothing wrong with it.

pranshugupta54:
It's not really important to check if instance is running or not, this is just setup!! 

Instance may or may not be running when user runs the actual server 😂

meetulr:
> It's not really important to check if instance is running or not, this is just setup!!

Now this is just flipping the whole thing😂. It is for checking a running instance, try running it without a running instance, you would be stuck in a loop.

> Instance may or may not be running when user runs the actual server 😂

We're checking for running instances, if not, we wouldn't be able to proceed. this is what the check is doing [link](https://github.com/PalisadoesFoundation/talawa-api/blob/7bd0b3f5c5fab60d705567a5a06a3b2f57006278/setup.ts#L426)

meetulr:
These are the confusions😂 maybe a clear description is need as to what it is meant to do.

After these discussions, my understanding is this:
That we're just checking whether the user is able to establish a connection with the provided mongo url, whether they actually mean to run the app right after, or tomorrow.
If that's the case, then just checking when the url changes makes sense, and the only things to consider/change here are the name/comment fixes, and that hard-coded url.

pranshugupta54:
I've made a flow on how everything will work after this PR is merged. 

We do NOT need `default: process.env.MONGO_DB_URL ?? "mongodb://localhost:27017"`.
_Assuming that we are having beginner developers,_
If the user added some random value in their env, suppose `MONGO_DB_URL=localhost:4000` (VERY WRONG). So our setup will first try to connect -> then fail -> prompt to ask new value (but default will show `localhost:4000` - very wrong). Now if we only keep `default: mongodb://localhost:27017`, then whatever messy value is being added by the user - they'll be suggested to use `mongodb://localhost:27017`


The following is how setup will work:
image

Apart from this, we need to add some more comments for each function.

Everything in this PR looks fine except some function naming and comments.
(Already discussed with @meetulr on Slack) - if someone else has anything else to say then please let me know

@palisadoes / @Anubhav-2003

Anubhav-2003:
> We do NOT need default: process.env.MONGO_DB_URL ?? "mongodb://localhost:27017".
Assuming that we are having beginner developers,
If the user added some random value in their env, suppose MONGO_DB_URL=localhost:4000 (VERY WRONG). So our setup will first try to connect -> then fail -> prompt to ask new value (but default will show localhost:4000 - very wrong). Now if we only keep default: mongodb://localhost:27017, then whatever messy value is being added by the user - they'll be suggested to use mongodb://localhost:27017
>
This seems to be a valid point. Suggesting new developers and first time comers the exact value of the URL, which is probably used 90% of times in localhost, will aid them in the setup as many people use mongoDB for the first time. 

The flowchart is also correct. The PR looks good. But one question, why is the event field changed in the schema in this pull request?  Is that relevant to this issue?

varshith257:
> But one question, why is the event field changed in the schema in this pull request?  Is that relevant to this issue?

It is the recent work of @meetulr to remove event status. I think it added into this PR of pulling with latest upstream in Pranshu local branch.

pranshugupta54:
It got automatically added by husky - pre-commit -> `gen:schema`

meetulr:
> It is the recent work of @meetulr to remove event status. I think it added into this PR of pulling with latest upstream in Pranshu local branch.

Yeah sorry about that😅, schema file didn't get updated in pr because i was having weird git issues  so i made a commit with --no-verify, and forgot to check it later😅

varshith257:
And one thing I like to add, A long back we had a PR merged to check the replicaSet or not and we have standardised it for a reason and me, @meetulr and @palisadoes are reviewed them. But it didn't introduced a way to use it except checking for replicaSet or not. Why don't we standardise this check in setup only to connect for a replicaSet?

palisadoes:
What's the status on this? It's not clear whether there approval for this PR or not

pranshugupta54:
@palisadoes, the flow is fine. I've added comments for functions which looked unclear.

Check https://github.com/PalisadoesFoundation/talawa-api/pull/2257#issuecomment-2074993217

palisadoes:
@varshith257 @meetulr @Anubhav-2003 PTAL once more

meetulr:
> @varshith257 @meetulr @Anubhav-2003 PTAL once more

I still have my doubts about variable naming. But others agree that these changes are good. We can see and change them later if people get confused.

meetulr:
> Everything seems fine but what about this?

@varshith257 do you mean to make replica set mandatory?

> But it didn't introduced a way to use it, except checking for replicaSet or not.

The way that check worked is that if we connect to a replica set, session would be initiated and we'll have transactions support, if not, session is not initiated and functionalities would still work.

If making replica set mandatory is what's required, it should have its own issue?

Anubhav-2003:
>If making replica set mandatory is what's required, it should be its own separate issue?
>
I agree on this point. This could be handled in a separate issue which I think  @varshith257 you can create. As for this PR. Everything seems fine including the flowchart.

Anubhav-2003:
> > @varshith257 @meetulr @Anubhav-2003 PTAL once more
> 
> I still have my doubts about variable naming😅. But others agree that these changes are good. We can see and change them later if people get confused.

If you have some doubts then you can suggest them in the review here. Then we can workout something all together.

pranshugupta54:
Most of the setup file is using similar variable name, so changing one of mongodb variable would require to change redis and more too. So I've simply added comments in a better way to make them easy. 

We can take replicaSet in a separate issue, I'm not very sure about how it works.

palisadoes:
Before merging. Will this allow setup to work with a local Mongo installation either:

1. Using docker?
2. Not using docker?

We had problems with this before. We need functionality in all scenarios.

Issue #2385, opened 2024-07-07
Feature Request


AVtheking:
**Is your feature request related to a problem? Please describe.**
Creating a installation script to simplify the installation process while setup  of the project

**Describe the solution you'd like**
Creating a nix flake to one click install the packages required for the setup of the project.

**Describe alternatives you've considered**
Creating a bash script for this but nix simplifies the installation process.

**Potential internship candidates**
Please read this if you are planning to apply for a Palisadoes Foundation internship https://github.com/PalisadoesFoundation/talawa/issues/359

RSN601KRI:
Hey @AVtheking 

The feature request in front of me is to create an installation script to make the process of project setup easier. I am ready to take on this task! Instead of bash-script, I think we could easily take a Nix flake approach to make installation a whole lot easier and very robust and reliable environment. The dependencies of packages by Nix are very easy to deal with so the setup would just get much easier. Let me know if you'd be okay with me adding to this, and I'll get this started immediately! 🙌

Thanks!

PurnenduMIshra129th:
Hi all. I am working on this issue previously 2 days . And already solved it . Using dockor now it the data are importing to mongo image database . i have to make one video about it then i will post here.

abhassen44:
can you assign me this issue if anyone is not working

Cioppolo14:
@AVtheking Is this resolved?

PR #2429, opened 2024-08-02
Devcontainer

AVtheking:
What kind of change does this PR introduce?

Created a devcontainer configuration to build the projects using devpods

Issue Number:

Fixes https://github.com/PalisadoesFoundation/talawa-api/issues/2385

Did you add tests for your changes?

No

If relevant, did you update the documentation?

Summary
Created configuration for the devpod to open the project in the prebuilt development environment.

Does this PR introduce a breaking change?
No

Have you read the [contributing guide](https://github.com/PalisadoesFoundation/talawa-api/blob/master/CONTRIBUTING.md)?

## Summary by CodeRabbit

- **New Features**
	- Added a new section in the installation guide for setting up the Talawa-API server using Devpod, with detailed instructions for both CLI and GUI versions.
	- Enhanced the Table of Contents for easier navigation to the new installation content.

- **Bug Fixes**
	- Simplified pre-commit script to focus on TypeScript type checks.


Dockerfile
```diff
@@ -0,0 +1,44 @@
+# Stage 1: Install Dependencies
```
varshith257, line 1: Why new Docker configuration? I think we can use existing Dockerfile.dev for it?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Stage 1: Install Dependencies
+FROM ubuntu:latest as builder
```
xoldd, line 2: use bookworm lts image

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN apt-get update && \
+    apt-get install -y git curl && \
```
xoldd, line 5: install node 20 lts

xoldd (commented):
I think this is decent for the first iteration and collecting feedback from other contributers. Add the documentation for working with devcontainers for contributers to refer to. After the changes can you make a short recording for the workflow and upload it to this pull request? That'd make things easier.

I have noticed certain issues, some of which coderabbitai bot has also mentioned, but I guess they're not as relevant right now.

INSTALLATION.md
```diff
@@ -288,6 +289,17 @@ We have created a setup script to make configuring Talawa-API easier.
 ```
 npm run setup
 ```
+## Install Using Devpod
+This guide provide step by step guide to setup a Talawa-Api server using devpod.
+
+Here is a list of steps to follow:
+1. Install the devpod gui application or devpod cli [Link](https://devpod.sh/docs/getting-started/install)
+2. Add a provider to the devpod , for [cli](https://devpod.sh/docs/getting-started/quickstart-devpod-cli#add-a-provider) and for [gui app](https://devpod.sh/docs/getting-started/quickstart-vscode#add-a-provider), You can add any provider from the list according to your needs, (we generally recommend using docker for this as this will host it on your local system , you can install docker from their official [documentation](https://docs.docker.com/engine/install/) follow this to install docker and add it as a provicder).
```
xoldd, line 297: Tell them to use docker or a docker compatible provider like podman, colima(better compatibility with macos) etc.

INSTALLATION.md
```diff
@@ -288,6 +289,17 @@ We have created a setup script to make configuring Talawa-API easier.
 ```
 npm run setup
 ```
+## Install Using Devpod
+This guide provide step by step guide to setup a Talawa-Api server using devpod.
+
+Here is a list of steps to follow:
+1. Install the devpod gui application or devpod cli [Link](https://devpod.sh/docs/getting-started/install)
+2. Add a provider to the devpod , for [cli](https://devpod.sh/docs/getting-started/quickstart-devpod-cli#add-a-provider) and for [gui app](https://devpod.sh/docs/getting-started/quickstart-vscode#add-a-provider), You can add any provider from the list according to your needs, (we generally recommend using docker for this as this will host it on your local system , you can install docker from their official [documentation](https://docs.docker.com/engine/install/) follow this to install docker and add it as a provicder). 
+3. Then just run command ```devpod up https://github.com/PalisadoesFoundation/talawa-api@pull/2429/head``` or follow this for [cli]( https://devpod.sh/docs/developing-in-workspaces/create-a-workspace#git-repository) and for [gui app](https://devpod.sh/docs/getting-started/quickstart-vscode#start-a-workspace-with-vs-code)
```
xoldd, line 298: Shouldn't the repo uri be `https://github.com/PalisadoesFoundation/talawa-api@develop`? I don't know the exact syntax but you have to target the `develop` branch within the uri.

AVtheking: Oh yeah, I did this for @palisadoes  to test it in this PR only

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "talawa api dev environment",
+  "dockerComposeFile": "docker-compose.yaml",
```
xoldd, line 3: shouldn't you utilize the already existing `docker-compose.dev.yaml` file?

xoldd (commented):
The core idea behind devcontainers is to bring development environment as close to production as is feasible if the project is making using of containers for its workflows. Read this:- https://12factor.net/dev-prod-parity

Reproducibility is a core requirement for detecting bugs easily. As such the base dependencies like the linux distribution and its exact version, node.js and its exact version and package managers like npm/yarn/pnpm and their exact version etc. must be the same in development and production environment. I would suggest debian bookworm as the linux distribution because it has better compatibility with the wider software ecosystem compared to alpine linux, for node.js I would suggest using latest lts version(20.17.0 as of writing this comment), npm comes pre-bundled with node.js so that isn't required.

Same goes for the services used within the docker compose file.

Now on top of that you can add additional stuff required in development environment that aren't required in the production like git, vim, neovim, github cli etc., as these tools are required by developers to work on the project within the container . It is better to strictly version these dependencies for reproducibility as well but it isn't a hard requirement.

.devcontainer/devcontainer.json
AVtheking: When I was using the compose file outside the  .decontainer.json file it was having some issues related to the project directory, that's why I used the file inside it. If I move the already existing file to that folder, developers can have issues, if they use docker for db.

palisadoes:
Is this ready to be reviewed?

AVtheking:
> Is this ready to be reviewed?

yes

INSTALLATION.md
```diff
@@ -288,6 +289,17 @@ We have created a setup script to make configuring Talawa-API easier.
 ```
 npm run setup
 ```
+## Install Using Devpod
+This guide provides a step-by-step guide to setting up a Talawa-Api server using Devpod.
+
+Follow these steps:
+
+1. Install the Devpod GUI application or Devpod CLI. [Learn more](https://devpod.sh/docs/getting-started/install)
```
palisadoes, line 297: Is this method independent of installing node.js and typescript as prerequisites? If it is, then it should be its own section. Think of the end user, not everyone is familiar with this technology.

INSTALLATION.md
```diff
@@ -288,6 +289,17 @@ We have created a setup script to make configuring Talawa-API easier.
...
+2. Add a provider to Devpod. For CLI, see [here](https://devpod.sh/docs/getting-started/quickstart-devpod-cli#add-a-provider) and for the GUI app, see [here](https://devpod.sh/docs/getting-started/quickstart-vscode#add-a-provider). Use Docker which you can install from their [official documentation](https://docs.docker.com/engine/install/) or  docker compatible provider like podman, colima(better compatibility with macos).
```
palisadoes, line 298: 1. Be more specific with the steps. Give the full list of steps if using the CLI and the GUI, or select one or the other. Whatever the choice, the steps must be foolproof. 
2. Don't give options, make the choice for the user. We don't have tech support, to the ease of use is very important

INSTALLATION.md
```diff
@@ -288,6 +289,17 @@ We have created a setup script to make configuring Talawa-API easier.
...
+3. Run the following command: ```devpod up https://github.com/PalisadoesFoundation/talawa-api@develop```, or follow the instructions for CLI [here](https://devpod.sh/docs/developing-in-workspaces/create-a-workspace#git-repository) and for the GUI app [here](https://devpod.sh/docs/getting-started/quickstart-vscode#start-a-workspace-with-vs-code).
```
palisadoes, line 299: 1. I am using the DevPod App image
2. I don't see instructions on how to get the CLI working.
3. The DevPod App image refers to workspaces which I assume I should be using. But it is not clear how

INSTALLATION.md
```diff
@@ -288,6 +289,17 @@ We have created a setup script to make configuring Talawa-API easier.
...
+4. Select your desired IDE.
```
palisadoes, line 300: If the user isn't a developer then they won't have an IDE. How will this help? Is there a solution to non-technical users to get the app functional?

More work needs to be done on the instructions. I need to be able to follow the instructions without having to know insider steps

palisadoes (changes requested):
More work needs to be done on the instructions. I need to be able to follow the instructions without having to know insider steps.

It is promising work. I hope this will make installation much easier and faster.

INSTALLATION.md
AVtheking: yes it is independent of installing node. js and ts , I didn't get with the point it should be its own section , could you clearify it more ?

INSTALLATION.md
AVtheking: we can have the api running without using ide , but for that it require little bit of change, I did just not spin up the api server using this is for developer as they may want to customize it according to their need.

INSTALLATION.md
palisadoes: We have a section in the table of contents for installing prerequisites. If we need to install different prerequisites depending on whether we use devcontainer or not, then we need the prerequisites section to be split in two for each scenario.
 
image

INSTALLATION.md
palisadoes: It needs to be clearer. You are assuming the only user will be a developer.

palisadoes:
1. Even with DevContainer you have to do setup, so it is not the desired one click ease of use.
2. I noticed that DevContainer has an AppImage. Could we use something similar to get to a one click solution?

INSTALLATION.md
AVtheking: See now

varshith257:
I see that both Devcontainer and Devpod are being included in the setup. While both tools have their strengths, could you clarify the specific use cases for each? Devcontainer already provides a consistent local environment, so I am curious about the rationale for including Devpod as well. Simplifying the setup might help reduce complexity for new developers.

INSTALLATION.md
AVtheking: I would suggest first let us test it for developers, then we can move to the users ease.

AVtheking:
> 1. Even with DevContainer you have to do setup, so it is not the desired one click ease of use.
> 
>     2. I noticed that DevContainer has an AppImage. Could we use something similar to get to a one click solution?

Do  you mean to create a AppImage like thing for the api server ?

AVtheking:
> I see that both Devcontainer and Devpod are included in the PR. While both tools have their strengths, could you clarify the specific use cases for each? Devcontainer already provides a consistent local environment, so I am curious about the rationale for including Devpod as well. Simplifying the setup might help reduce complexity for new developers.
> 
> If the focus is on achieving a true one-click setup experience for local development, Devpod may not be necessary.
> 
> if remote capabilities are necessary, consider using Devpod as the standalone setup tool with DevContainer as an optional setup.

see Devpod uses devContainer.json for its underlying implementation , I agree we can achieve this with the help of devcontainer but it is specific to vs code only , with devpod we can create this environment in any ide or without any ide.

AVtheking:
> 1. Even with DevContainer you have to do setup, so it is not the desired one click ease of use.
> 
>     2. I noticed that DevContainer has an AppImage. Could we use something similar to get to a one click solution?

by the way , devpod requires just initial setup that is installing of devpod setting up of provider.... This would still be required by the user to host the server , with devpod it is just one command and the server would spin up with all the dependencies it requires.

palisadoes:
@AVtheking Can we merge this now?

AVtheking:
> @AVtheking Can we merge this now?

yes lets start with the initial version and then we can iterate based on the users feedback.

PR #3239, opened 2025-02-17
Fixed Backward Pagination

### LINKED ISSUE #3237: Bug: Faulty Backward Pagination
**Describe the bug**
Currently the backward pagination is faulty. This issue aims to fix this.

**To Reproduce**

1. Try Backward Pagination using graphiql

**Expected behavior**
The backward pagination should work seamlessly as the forward pagination given the startCursor and endCursor.

**Actual behavior**
The issue was with the way we handled the previousPage check. We were fetching limit + 1 items from the database to determine if a previous page existed. If the length of the fetched data was equal to limit + 1, it indicated that a previous page was available. After setting the previousPage flag to true, we were supposed to remove the extra fetched element (i.e., the last item), but the existing logic was incorrectly removing the first element instead.

This caused backward pagination to skip the last item and instead return elements beyond the intended range. As a result, the pagination would break—forward pagination worked fine, but backward pagination was returning incorrect sequences.

**Screenshots**

https://github.com/user-attachments/assets/78601f1b-8476-4e9a-8a73-729945d8f737

Please read this if you are planning to apply for a Palisadoes Foundation internship 
- https://github.com/PalisadoesFoundation/talawa/issues/359

rahulch07:

This PR fixed the bug in backward pagination

**Issue Number:**

Fixes #3237 

**Snapshots/Videos:**

[Pagination Bug Fixed.webm](https://github.com/user-attachments/assets/9afd2b51-996e-4391-9acf-be2ab8fa8776)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pagination behavior to ensure consistent page navigation and accurate boundary determination.
- **Tests**
  - Added robust test suites to validate pagination transformations and confirm reliable performance.


palisadoes (changes requested):
Please add tests for this file so that the issue doesn't happen again. Try to get close to 100% code coverage

palisadoes:
It's good to see that this is fixed, but it would have been caught with better testing. It's only 50% covered now. Please add the appropriate tests

image

rahulch07:
> 2\. target

Okay adding test cases

palisadoes:
1. Please make coderabbit.ai approves your work.
2. Make sure all tests pass and are valid.
3. Ensure the test code coverage for your patch reaches as close to 100% as possible.

palisadoes (approved):

Issue #3214, opened 2025-02-14
NoMongo: GitHub action to test basic DB access and updates


palisadoes:
**Is your feature request related to a problem? Please describe.**

Our GitHub actions test basic functionality, but don't test key functionalities on live data.

**Describe the solution you'd like**

Create GitHub actions that test the importation of sample data

1. Validate the existence of data in all tables
1. Update various tables through representative API calls and validate the results

Depends on:
 
- https://github.com/PalisadoesFoundation/talawa-api/pull/2952

**Describe alternatives you've considered**

- N/A

**Approach to be followed (optional)**

- See above

**Additional context**

Depends on:
 
- https://github.com/PalisadoesFoundation/talawa-api/pull/2952

**Potential internship candidates**

Please read this if you are planning to apply for a Palisadoes Foundation internship 
- https://github.com/PalisadoesFoundation/talawa/issues/359

JaiPannu-IITI:
@palisadoes please assign

PurnenduMIshra129th:
@palisadoes  please assign.

JaiPannu-IITI:
Thankyou @dhanagopu

palisadoes:
@JaiPannu-IITI 

When the containers start the `src/plugins/seedInitialData.ts` file automatically inserts a `API_ADMINISTRATOR_USER_EMAIL_ADDRESS` user in the DB if none already exists. That is probably the source of the duplicate entry. The sample DB needs to be updated, not the script.

JaiPannu-IITI:
@palisadoes No sir

Every user should have unique user_id, if we try to use sample data script two times, it was conflicting earlier due to this, I have fixed that thing, now it will no throw error on this.

sample data script is never designed to add API_ADMINISTRATOR_USER in the first place, it was other data that was conflicting which has been fixed,

JaiPannu-IITI:
Sample data script will not run if API_ADMINISTRATOR_USER is not present, that's why I used seedInitialData helper function here to deal with cases when db is reset intentionally and then user tries to add sample data, then administrator will be recreated, sample data has nothing to do with administrator

I strongly recommend you to try executing sample data and reset scripts in all cases, you'll find out whatever user does like format db, change administrator or change sample data anything, they will not throw error and fulfill their purpose

JaiPannu-IITI:
https://github.com/PalisadoesFoundation/talawa-api/issues/3243

Sir this issue is not related to this, this issue is long existing but reported just now

[...]

PR #3304, opened 2025-02-26
Minor Update -> reset:data

JaiPannu-IITI:
### Overview

Based on discussion about testing the scripts with xoldd, it came into light that we don't need to reset administrator and recreate (similar to what seedInitialData does), we should ignore this since it remains unaffected from users. 

This PR introduces minor change in script to skip deleting administrator from DB and skip checking or recreation of administrator in both reset:data or add:sample_data since it is absolute that administrator always exist in any situation even after a db reset.

Related Issue: #3214 

### Future Steps
Since we are close to resolving issue regarding conflicts that might occur during parallel vitest, this PR will be followed by adding new tests and correcting old ones to avoid conflict while running in parallel.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined database initialization and cleanup by removing explicit administrator setup checks.
	- Adjusted logging to now simply confirm that administrator access is preserved after database operations.
- **Tests**
	- Updated test cases to align with the simplified database processes, eliminating scenarios related to administrator verification.

### DIFF
--- a/scripts/dbManagement/addSampleData.ts
+++ b/scripts/dbManagement/addSampleData.ts
@@ -1,11 +1,6 @@
 import path from "node:path";
 import { fileURLToPath } from "node:url";
-import {
-	disconnect,
-	ensureAdministratorExists,
-	insertCollections,
-	pingDB,
-} from "./helpers";
+import { disconnect, insertCollections, pingDB } from "./helpers";
 
 type Collection =
 	| "users"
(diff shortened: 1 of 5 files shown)


JaiPannu-IITI:
@palisadoes Please ignore this Patch for now, it is related to helpers file for which tests are not written right now, I am working on that. 

PR is ready for review.

palisadoes (approved):

Issue #3569, opened 2025-09-07
Race Condition in Event Registration Allows Overbooking Beyond Capacity


vamshi567bh:
When multiple users attempt to register for the same event simultaneously, the API sometimes allows more registrations than the event's defined capacity. This results in overbooking, which can cause issues for event organizers and participants.

To Reproduce:

Create an event with a limited capacity (e.g., 10 seats).
Simultaneously send multiple registration requests (using scripts or tools like JMeter/Postman) from different users when only a few seats remain.
Observe that the total number of registered users can exceed the event's capacity.
Expected behavior: The API should enforce the event capacity strictly, preventing more users from registering than allowed, even under high concurrency.

Actual behavior: Due to a race condition, the API sometimes allows more registrations than the event's capacity.

Additional details: This bug likely occurs because the check for available seats and the registration update are not performed atomically. Implementing proper locking or transactional logic is required to fix this issue.

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
@gautam-divyanshu please assign me

vamshi567bh:
@palisadoes  assign me

vamshi567bh:
@tasneemkoushar please assign me

vamshi567bh:
@DMills27  please assign me

vamshi567bh:
@varshith257 please assign me

palisadoes:
@BharathC0

You already have >=2 issues assigned to you. Our policy is 2 issues across all repositories. When you have merged the PRs related to your issues, then we can consider assignments. 

Please refrain from asking to be assigned. You have requested to be assigned to dozens of issues that requires us to look at the email, validate the number of issues assigned and then leave this message. Please be considerate of our time.

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
@varshith257 please assign me

vamshi567bh:
@varshith257 please assign me

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
@palisadoes please review

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
@palisadoes please assign me

vamshi567bh:
[...]

PR #3607, opened 2025-10-05
Fix: Atomic Event Registration and EventAttendance Type #3569

BharathC0:
Implements atomic event registration logic in registerForEvent.ts to strictly enforce event capacity and prevent race conditions during concurrent registrations.
Defines and exports the EventAttendance GraphQL object type for correct mutation output and schema consistency.
Updates import paths to use relative references for compatibility.
Applies Biome formatting and code quality standards to both files.     fixes: #3569


BharathC0:
@coderabbitai what is the problem for codecov failing issue

PR #3611, opened 2025-10-08
enforce strict event capacity under concurrency and resolve type errors #3569

### LINKED ISSUE #3569: Race Condition in Event Registration Allows Overbooking Beyond Capacity

BharathC0:
Ensures the API strictly enforces event capacity, preventing overbooking even when multiple users register simultaneously.
Refactored registration logic to use atomic database transactions for capacity checks and inserts.
Resolves all Biome and TypeScript type errors, removes explicit any usages, and improves type safety.
Steps to reproduce the original issue:
Create an event with limited capacity (e.g., 10 seats).
Simultaneously send multiple registration requests from different users.
Previously, registrations could exceed capacity; now, the API prevents this.
All changes pass lint and type checks.
Ready for review and merge into develop.   fixes #3569


vamshi567bh:
@coderabbit which are failing tests and files

